PR #25, opened 2016-08-26
Draft Tutorial.md

tbillington:


mish15:
@tbillington is this still valid?

tbillington:
@mish15 It's quite of out date and doesn't contribute a whole lot, I'll close.

PR #59, opened 2017-07-23
Modeling new state

dhowden:
Annoying Github doesn't let you review your own pull request?


src/ui/index.js
```diff
@@ -1 +1,5 @@
-export { BodyInput } from './BodyInput.js'
+import * as facets from "./facets";
+import * as overlay from "./overlay";
+import * as results from "./results";
+import * as searchInput from "./searchInput";
```
dhowden: `text`?

src/ui/facets/index.js
```diff
@@ -0,0 +1,6 @@
+import * as Debug from "./Debug";
+import * as Input from "./Input";
+
+export { default as TabsFacet } from "./TabsFacet";
+
+export { Debug, Input };
```
dhowden: Fix this to be like the rest, seems weird to be different? (want `import { RadioFacet } from `sajari-react/ui/facets`)

src/ui/facets/TabsFacet.js
```diff
@@ -0,0 +1,59 @@
+import React from 'react'
+import PropTypes from "prop-types";
+
+import Values from "../../controllers/values";
+import Pipeline from "../../controllers/pipeline";
+import Filter from "../../controllers/filter";
+
+class TabsFacet extends React.Component {
```
dhowden: This needs to use `singleFacet`

src/controllers/pipeline.js
```diff
@@ -0,0 +1,126 @@
...
+  trackingResetEvent,
+];
+
+class Pipeline {
+  constructor(client, name, values, tracking) {
+    this.client = client;
+    this.name = name;
+    this.values = values;
+    this.tracking = tracking;
+    this.afterValuesChanged = this.values.listen(postChangeEvent, this.afterValuesChanged.bind(this));
```
dhowden: Is this being used?

src/controllers/multiFacet.js
```diff
@@ -0,0 +1,45 @@
...
+
+  notify() {
+    this.listener.notify(l => {
+      l(this);
+    });
+  }
+
+  set(name, on) {
+    if (on) {
+      this.current = this.current.concat(name);
```
dhowden: Check that it's not already on first?

src/controllers/filter.js
```diff
@@ -0,0 +1,57 @@
+const andFilter = "AND";
+const orFilter = "OR";
+const singleFilter = "single";
+
+export const ANDFilter = () => new Filter(andFilter);
+export const ORFilter = () => new Filter(orFilter);
+
+export class Filter {
```
dhowden: This now looks a lot like `MultiFacet`

src/analytics/analytics.js
```diff
@@ -1,21 +1,17 @@
...
-  State,
-  RESULTS_CHANGED,
-  TRACKING_RESET,
-  RESULT_CLICKED
-} from "./state";
+import { resultsEvent, trackingResetEvent, resultClickedEvent } from "../controllers/pipeline";
 
 class Analytics {
-  constructor(namespace, ga = new GA()) {
+  constructor(pipeline, values, ga = new GA()) {
```
dhowden: Need to remove the need to pass values here.

src/analytics/analytics.js
```diff
@@ -57,13 +53,13 @@ class Analytics {
...
   onChange() {
-    const searchResponse = State.ns(this.namespace).getResults();
+    const searchResponse = this.pipeline.getResults();
     // Enable analytics once a successful search has been performed
     if (searchResponse && searchResponse.results) {
       this.enabled = true;
     }
 
-    const values = State.ns(this.namespace).getValues();
+    const values = this.values.get();
```
dhowden: Ok, so I think you can just get the values from the pipeline? (i.e. in `Pipeline` store a "flat" copy of the values dictionary used in the `Search` request).

examples/standard-site-integration/src/stateProxy.js
```diff
@@ -0,0 +1,14 @@
+import { values, pipeline } from "./resources";
+
+class StateProxy {
```
dhowden: `ValuesProxy`? Maybe this is obsolete now we have `Values` and `Pipeline`?

examples/standard-site-integration/src/index.js
```diff
@@ -135,12 +153,24 @@ const initInterface = () => {
...
-  _state.setCollection(config.collection);
-  _state.setPipeline(config.pipeline);
+  initialiseResources(config.project, config.collection, config.pipeline)
 
   if (!config.disableGA) {
-    new Analytics("default");
+    new Analytics(pipeline, values);
+  }
+
+  if (config.tabFilters && config.tabFilters.defaultTab) {
```
dhowden: This should disappear by using `SingleFacet` to control the tabs.

examples/standard-site-integration/src/index.js
```diff
@@ -119,8 +128,17 @@ const initInPage = config => {
     config.attachSearchResponse
   );
 
-  const values = combinedValues(config, true);
-  _state.setValues(values, values.q);
+  const queryValues = combinedValues(config, true);
+  if (queryValues.filter) {
+    filter.setFilter("dictionary", queryValues.filter);
```
dhowden: `"dictionary"`?  `"initialValues"` or something more informative instead.

examples/standard-site-integration/src/index.js
```diff
@@ -72,12 +74,15 @@ const initOverlay = config => {
   const setOverlayControls = controls => {
     const show = () => {
       document.getElementsByTagName("body")[0].style.overflow = "hidden";
-      _state.setValues(combinedValues(config, false));
+
+      // _state.setValues(combinedValues(config, false));
```
dhowden: Remove

examples/standard-site-integration/src/index.js
```diff
@@ -72,12 +74,15 @@ const initOverlay = config => {
...
       controls.show();
     };
     const hide = () => {
       document.getElementsByTagName("body")[0].style.overflow = "";
-      _state.reset();
+      // _state.reset();
```
dhowden: Remove

examples/standard-site-integration/src/index.js
```diff
@@ -45,24 +48,23 @@ const checkConfig = () => {
...
 const combinedValues = (config, firstTime) => {
   let initialValues = {};
-  // Only include initial values the first time App is initialise0d
+  // Only include initial values the first time App is initialised
   if (config.initialValues && firstTime) {
     initialValues = config.initialValues;
   }
 
-  const tabValues = {};
+  // const tabValues = {};
```
dhowden: Remove

examples/standard-site-integration/src/index.js
```diff
@@ -1,20 +1,23 @@
...
+import { changeEvent } from "sajari-react/controllers/values";
+
+import { initialiseResources, pipeline, values, filter } from "./resources";
+
 import "./styles.css";
 
 const ESCAPE_KEY_CODE = 27;
 
-const _state = State.default();
+// const _state = State.default();
```
dhowden: Remove

src/controllers/pipeline.js
```diff
@@ -0,0 +1,126 @@
...
+const events = [
+  searchEvent,
+  errorEvent,
+  resultsEvent,
+  resultClickedEvent,
+  trackingResetEvent,
+];
+
+class Pipeline {
+  constructor(client, name, values, tracking) {
```
dhowden: Move `values` to `Search`?

src/ui/text/index.js
```diff
@@ -0,0 +1 @@
+export { default as AutocompleteInput } from "./AutocompleteInput";
```
tbillington: Added

src/ui/overlay/index.js
```diff
@@ -0,0 +1,3 @@
+export { default as Overlay } from "./Overlay";
+export { default as Close } from "./Close";
+export { default as Logo } from "./Logo";
```
tbillington: Added

examples/autocomplete-only/.gitignore
```diff
@@ -0,0 +1,21 @@
...
+/node_modules
+
+# testing
+/coverage
+
+# production
+/build
+
+# misc
+.DS_Store
```
jonathaningram: @tbillington I just saw you have committed .DS_Store files above?

dhowden: @tbillington We should all have .DS_Store in our global .git-ignore files...

package.json
```diff
@@ -23,6 +23,7 @@
     }
   ],
   "dependencies": {
```
jonathaningram: Using `prettier` yet?

src/ui/facets/Debug.js
```diff
@@ -0,0 +1,48 @@
...
+import PropTypes from "prop-types";
+
+import { Filter } from "../../controllers/filter";
+
+class DebugFacet extends React.Component {
+  /**
+   * propTypes
+   * @property {Filter} filter Filter instance to debug.
+   */
+  static get propTypes() {
```
jonathaningram: Should be able to just do:

```js
static propTypes = { filter: PropTypes.instanceOf(Filter).isRequired }
```

src/ui/facets/Input.js
```diff
@@ -0,0 +1,123 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+import { Filter } from "../../controllers/filter";
+
+class CheckboxFacet extends React.Component {
```
jonathaningram: This can be a dumb component:

```js
const CheckboxFacet = props => <InputFacet ... />;

CheckboxFacet.propTypes = {...};
```

src/ui/facets/Input.js
```diff
@@ -0,0 +1,123 @@
...
+    this.unregister();
+  }
+
+  onUpdate(filter) {
+    this.setState({ active: filter.get() });
+  }
+
+  render() {
+    const { filter, name, options, ...other } = this.props;
+    const onClick = e => {
```
jonathaningram: Don't create new functions in `render`.

src/ui/facets/Input.js
```diff
@@ -0,0 +1,123 @@
...
+      </select>
+    );
+  }
+}
+
+class InputFacet extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { active: props.filter.isSet(props.name) };
+    this.onUpdate = this.onUpdate.bind(this);
```
jonathaningram: Use autobinding methods instead of binding in constructor.

src/ui/facets/Input.js
```diff
@@ -0,0 +1,123 @@
...
+  componentWillUnmount() {
+    this.unregister();
+  }
+
+  onUpdate(filter) {
+    this.setState({ active: filter.isSet(this.props.name) });
+  }
+
+  render() {
+    const { filter, name, children, ...other } = this.props;
```
jonathaningram: Call it `rest` - is the convention.

src/ui/facets/Tabs.js
```diff
@@ -0,0 +1,77 @@
...
+   */
+  static get propTypes() {
+    return {
+      filter: PropTypes.oneOfType([PropTypes.instanceOf(Filter)]).isRequired,
+      tabs: PropTypes.array.isRequired,
+      defaultTab: PropTypes.string.isRequired
+    };
+  }
+
+  constructor(props) {
```
jonathaningram: Useless constructor.

src/ui/facets/Tabs.js
```diff
@@ -0,0 +1,77 @@
...
+    super(props);
+  }
+
+  render() {
+    return (
+      <div className="sj-tabs-container">
```
jonathaningram: If you're going to use self-written class names, best to use BEM or something.

src/ui/facets/Tabs.js
```diff
@@ -0,0 +1,77 @@
...
+        <div className="sj-tabs">
+          {this.props.tabs.map(t =>
+            <Tab key={t.title} title={t.title} filter={this.props.filter} />
```
jonathaningram: Is `t.title` unique? If not, use `i/index` from the map closure.

src/ui/facets/Tabs.js
```diff
@@ -0,0 +1,77 @@
...
+
+  onChange() {
+    this.setState({ selected: this.props.filter.isSet(this.props.title) });
+  }
+
+  render() {
+    return (
+      <div
+        key={this.props.title}
+        className={`sj-tab${this.state.selected ? " sj-tab-active" : ""}`}
```
jonathaningram: E.g. with BEM (if I remember the name correctly) I think the class name will be `sj-tab--active`.

src/ui/facets/Tabs.js
```diff
@@ -0,0 +1,77 @@
...
+        onClick={() => {
```
jonathaningram: Don't create new functions in `render`.

src/ui/results/Image.js
```diff
@@ -0,0 +1,23 @@
+import React from "react";
+
+class Image extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { show: true };
+  }
+
+  render() {
+    const onError = () => this.setState({ show: false });
```
jonathaningram: No new functions in `render`.

src/ui/results/Response.js
```diff
@@ -0,0 +1,68 @@
...
+    const error = pipeline.getError();
+
+    if (!time && !error) {
+      return Placeholder ? <Placeholder /> : null;
+    }
+
+    const propsForChildren = { ...results, error, pipeline };
+    const childrenWithResults = React.Children.map(children, c => {
+      if (React.isValidElement(c)) {
+        return React.cloneElement(c, propsForChildren);
```
jonathaningram: Usually a code smell if you are calling `cloneElement`.

src/ui/results/Url.js
```diff
@@ -0,0 +1,15 @@
+import React from "react";
+
+import TokenLink from "./TokenLink";
+
+const Url = ({ url, token, resultClicked }) =>
```
jonathaningram: `URL`?

src/ui/text/Input.js
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  getState(values, qParam) {
+    const text = values.get()[qParam] || "";
+    return { text };
+  }
+
+  componentDidMount() {
+    if (this.props.focus) {
+      findDOMNode(this.refs.searchInput).focus();
```
jonathaningram: Any reason you're not just setting the `autoFocus` attribute on the `<input />`? Shouldn't really be finding the DOM node and calling it?

src/ui/text/Input.js
```diff
@@ -0,0 +1,96 @@
...
+      this.valuesUpdated
+    );
+  }
+
+  componentWillUnmount() {
+    this.removeValuesListener();
+  }
+
+  valuesUpdated() {
+    this.setState(this.getState(this.props.values, this.state.qParam));
```
jonathaningram: This is confusing. React components use the term `state` and you've added `getState` but it makes me think it's calling the React component's state, but it's not. Mixing the two seems like a bad idea...

src/ui/text/Input.js
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  render() {
+    const { text } = this.state;
+    const { instant, ...other } = this.props;
+
+    return (
+      <input
+        type="text"
+        ref="searchInput"
```
jonathaningram: Shouldn't be using string `ref`s anymore.

src/ui/facets/Input.js
dhowden: This seems to have become our convention :/  Mass search-replace here-we-come!

src/ui/facets/Tabs.js
dhowden: So we were talking about this.  Suggestions for alternative to class names?  This code is being used in the auto-generated interfaces (which is where the classnames come from).  We can change these and bump the version (of course), so basically anything goes as a soln.

package.json
dhowden: Needs to be applied globally, I noticed a few things yesterday too.

src/ui/facets/Tabs.js
dhowden: Agreed :-P

src/ui/facets/Tabs.js
dhowden: If it's not unique, then the filter won't work :-P

src/ui/results/Response.js
dhowden: :-)

src/ui/text/Input.js
dhowden: Agreed.

src/ui/facets/Tabs.js
jonathaningram: Fair enough - except `key` is special and developers look out for it to mean the React key, so I'd just use `i` - otherwise looking at `key={t.title}` it looks like it has a more special function than it actually does `key={i}` says exactly that it's just a React thing.

src/ui/results/Response.js
```diff
@@ -0,0 +1,68 @@
...
+
+    return (
+      <div className="sj-pipeline-response">
+        {childrenWithResults}
+      </div>
+    );
+  }
+}
+
+Response.propTypes = {
```
jonathaningram: When you're not using dump components, I think it's a better convention to use `static propTypes = {};` and `static defaultProps = {};` within the class, rather than outside after the declaration.

src/ui/results/Response.js
```diff
@@ -0,0 +1,68 @@
...
+      this.onResultsChange
+    );
+  }
+
+  componentWillUnmount() {
+    this.removeErrorListener();
+    this.removeResultsListener();
+  }
+
+  onResultsChange() {
```
jonathaningram: This should be called `handleResultsChange`. See https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-handler-names.md - it's a good convention to follow as it lets us reason about event handler and what a method does just by the name. tl;dr - props are called `onX` and the value you assign to it is `handleX` e.g. `<C onChange={this.handleChange />`.

src/ui/results/Response.js
```diff
@@ -0,0 +1,68 @@
...
+    const response = this.props.pipeline.getResults() || {};
+    this.setState({ results: response });
+  }
+
+  render() {
+    const { children, Placeholder, pipeline } = this.props;
+    const results = this.state.results;
```
jonathaningram: `const { results } = this.state;`

src/ui/results/Response.js
```diff
@@ -0,0 +1,68 @@
...
+    const time = results.time;
+    const error = pipeline.getError();
+
+    if (!time && !error) {
```
jonathaningram: I don't understand this conditional? Can it be written to be more obvious? "If we don't have a time and we don't have an error return the placeholder if we have it otherwise null"?

src/ui/results/Response.js
jonathaningram: OK so `results` comes off the `pipeline` and is assigned to this component's state object in the constructor. Why not just pass `results` to each of the children one level up? E.g.

```js
<Response pipeline={pipeline}>
      <Summary values={values} pipeline={pipeline} tracking={tracking} results={pipeline.getResults()} />
      <Results pipeline={pipeline} results={pipeline.getResults()} />
      <Paginator values={values} pipeline={pipeline} tracking={tracking} results={pipeline.getResults()} />
    </Response>
```

I know that looks strange and I think there is still a better way, and I know that the `onResultsChange` is updating the state too, but yeah it's definitely not looking like the right way to do this.

src/ui/facets/Input.js
dhowden: Roger.

src/ui/facets/Input.js
dhowden: Roger.

src/ui/facets/Tabs.js
dhowden: Ack

src/ui/results/Image.js
dhowden: Ack

src/ui/text/Input.js
dhowden: ❤️

src/ui/text/Input.js
dhowden: Yup. Not necessary after `autoFocus` fix :-).

src/ui/facets/Tabs.js
tbillington: Don't have a strong opinion either way, was following the react docs which say to not use index as key unless you have no other option: https://facebook.github.io/react/docs/lists-and-keys.html#keys.

In this situation though I guess it wouldn't matter because tabs don't reorder during their lifetime.

examples/autocomplete-only/.gitignore
tbillington: Oops.. It's in global ignore now :)

PR #64, opened 2017-08-09
tracking: handle the case where values doesn't have a q value

tbillington:
Values not having a `q` value is a normal situation and should not cause the code to error.


tbillington:
Included in a more recent pull request.

PR #65, opened 2017-08-09
pipeline: add cancel search, use when clearing ui inputs

tbillington:
Currently, if a search request completes after an input component has been cleared the results will still be used.

This change will prevent any searches currently in progress from being used after `pipeline.cancelSearch()` has been called.


tbillington:
Closing this without merging as this should be happening somewhere else (pipeline/values/tracking)

Issue #62, opened 2017-08-01
Chars overlapping in search-box input


tolyadouble:
If you type a lot characters, fill the whole space, in the search input field, chars start overlap each other.

benhinchley:
closing due to #87

PR #67, opened 2017-08-17
design: fix default search box for mobile

mish15:
previously the input was 500px hardcoded, which causes issues on
mobile. It now assumes the full width of the parent container which
makes it easy to change and also works on mobile. Also adds some small
spacing and a slight border radius. Should help with #62 also.


tbillington:
Included in a more recent pull request.

PR #84, opened 2018-03-17
package: add esmodules build

benhinchley:


tbillington:
It looks like the test files are included in the build, I noticed this because when the tests ran I saw double the amount, half coming from the build directory 😛

benhinchley:
good catch didn't notice that before haha, should be fixed as of 36199ae

benhinchley:
closing due to #87

PR #87, opened 2018-06-25
React SDK v2

benhinchley:


src/components/Input/components/InputBox/styled.ts
```diff
@@ -0,0 +1,128 @@
...
+  "& input:focus, & input:active": {
+    outline: "none"
+  }
+});
+
+export const inputResetStyles = {
+  container: {
+    "& > div[style]": {
+      display: "inline",
+      // fixes sizing bug in firefox
```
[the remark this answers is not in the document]
benhinchley: FF V 60.0.2

src/components/Select/Select.tsx
```diff
@@ -0,0 +1,67 @@
...
+    })),
+  isEqual
+);
+
+const handleChange = (
+  selected: string[],
+  set: (name: string, value: boolean) => void
+) => (value: any, { action }: any) => {
+  switch (action) {
+    case "select-option":
```
dhowden: These should maybe be consts from somewhere? (most un-JS of me, I know)

src/components/Summary/Summary.tsx
```diff
@@ -0,0 +1,104 @@
...
+            response.isEmpty() ||
+            response.isError()
+          ) {
+            return null;
+          }
+          const responseValues = response.getValues() as Map<string, string>;
+          const queryValues = response.getQueryValues() as Map<string, string>;
+
+          const text = responseValues.get(config.qParam) || query;
+          const page = parseInt(queryValues.get("page") as string, 10);
```
dhowden: `"page"` should probably be `pageParam` (similar to `qParam`)

src/components/Paginator/Paginator.tsx
```diff
@@ -0,0 +1,140 @@
...
+          ) {
+            return null;
+          }
+
+          const queryValues = response.getQueryValues();
+          if (queryValues === undefined) {
+            return;
+          }
+
+          const page = queryValues.get("page")
```
dhowden: `"page"` -> `pageParam` (similar to how `qParam` is used).

src/components/Paginator/Paginator.tsx
```diff
@@ -0,0 +1,140 @@
...
+            ? parseInt(queryValues.get("page") as string, 10)
+            : 1;
+          const resultsPerPage = queryValues.get("resultsPerPage")
```
dhowden: `"resultsPerPage"` -> `resultsPerPageParam` (similar to how `qParam` is used).

src/components/context/pipeline/Provider.tsx
```diff
@@ -0,0 +1,293 @@
...
+      )
+    );
+
+    this.instant = instant;
+    if (this.instant === undefined) {
+      const { project, collection, endpoint } = search.pipeline.config;
+      this.instant = {
+        pipeline: new Pipeline(
+          { project, collection, endpoint },
+          "autocomplete",
```
dhowden: This should be a parameter I think?

src/components/context/pipeline/Provider.tsx
```diff
@@ -0,0 +1,293 @@
...
+    const { search, instant } = this.props;
+
+    this.setState(
+      state =>
+        ({
+          ...state,
+          search: {
+            ...state.search,
+            response: search.pipeline.getResponse(),
+            query: search.values.get().q || "",
```
dhowden: I think this should be a paramter?

src/components/Input/Input.tsx
```diff
@@ -0,0 +1,312 @@
...
+  UpArrow = 38,
+  RightArrow = 39,
+  DownArrow = 40
+}
+
+export type InputMode = "standard" | "typeahead";
+export type DropdownMode = "none" | "suggestions" | "results";
+
+export interface InputProps {
+  inputMode?: InputMode;
```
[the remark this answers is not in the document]
benhinchley: yeah can switch to `mode`

src/components/Input/components/InputBox/InputBox.tsx
```diff
@@ -0,0 +1,210 @@
...
+          </ButtonContainer>
+        </SearchContainer>
+      </InputContainer>
+    );
+  }
+
+  private handleSearchButtonOnClick = (event: ButtonMouseEvent) => {
+    const { onSearchButtonClick } = this.props;
+    event.preventDefault();
+    if (typeof onSearchButtonClick === "function") {
```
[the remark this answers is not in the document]
benhinchley: yeah basically it stops the check from being `(onSearchButtonClick !== null && 
onSearchButtonClick !== undefined && typeof onSearchButtonClick === "function")`,

The function check needs to be there to prevent the possibility of trying to call something that isn't a function.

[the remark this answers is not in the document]
dhowden: I think it would be useful for these methods to throw errors if they are set to some non-null thing that isn't a function (rather than just silently ignoring it).  Would catch a lot of things where I normally get errors like "X is an object, can't use it like a function" std JS error.

src/components/Select/Select.tsx
benhinchley: these technically are consts inside of react-select, i'll see if I can reference them directly from the module

src/docs/localization.mdx
```diff
@@ -0,0 +1,52 @@
...
+route: /i18n
+---
+
+# Localization
+
+Under the hood we are using [i18next](https://www.i18next.com/) to handle localisation
+of various parts of our search interfaces. We export our instance of [i18next](https://www.i18next.com/)
+so that it is possible to provide your own localisations for your search interfaces.
+
+By default we load the relevant localization dictionary based on the a users browser settings.
```
dhowden: `based on the user's browser settings.`

src/docs/localization.mdx
```diff
@@ -0,0 +1,52 @@
+---
+name: Localization
+route: /i18n
+---
+
+# Localization
+
+Under the hood we are using [i18next](https://www.i18next.com/) to handle localisation
```
dhowden: localization

src/components/context/pipeline/Provider.tsx
benhinchley: This is the one that is created if you do not provide an instant pipeline config

PR #89, opened 2018-09-13
various updates and fixes

benhinchley:
* The `<Input />` component has been rewritten to use [downshift](https://github.com/paypal/downshift) under the hood, which has solved a number of accessibility, stability issues with the component.
* the `dropdownMode` prop on the`<Input />` now accepts `results` instead of `custom`
* The `DropdownRenderer` on the `<Input />` has been removed
* The `<Overlay />` container now has different classes for the shade and the container -> `.sj-overlay__shade` and `.sj-overlay__container`


src/components/Input/Input.tsx
```diff
@@ -175,151 +135,455 @@ export class Input extends React.Component<InputProps> {
...
+                      }
+                      const url =
+                        Object.getOwnPropertyNames(item.token).length > 0
+                          ? item.token
+                          : item.values.url;
+                      window.location.href = url;
+                      return;
+                    }
+
+                    item = downshift.selectedItem;
```
benhinchley: this can be removed, already defaulting to this above

src/components/Input/Input.tsx
```diff
@@ -175,151 +135,455 @@ export class Input extends React.Component<InputProps> {
...
+        {this.props.dropdownMode === "suggestions" && (
+          <Suggestions
+            downshift={downshift}
+            typedInputValue={this.state.typedInputValue}
+            suggestions={items as string[]}
+            styles={this.props.styles && this.props.styles.suggestions}
+          />
+        )}
+        {this.props.dropdownMode === "results" && (
+          <Results downshift={downshift} />
```
benhinchley: We don't seem to be passing through the `ResultRenderer` prop through to this component.

PR #101, opened 2019-01-17
Map results view component

### LINKED ISSUE #99: Create map based search results view
Create map based search results view that can plot a geo-location associated with a search result and show it on a map e.g. Google Maps.

image

zlatanpham:
Resolve #99

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,6 +57,7 @@
     "@babel/preset-react": "^7.0.0",
     "@babel/preset-typescript": "^7.0.0",
     "@types/enzyme": "^3.1.13",
+    "@types/googlemaps": "^3.30.16",
     "@types/i18next": "^11.9.0",
     "@types/jest": "^23.3.2",
     "@types/react": "^16.4.14",
--- a/src/components/MapResultsView/DefaultMarker.tsx
+++ b/src/components/MapResultsView/DefaultMarker.tsx
@@ -1,0 +1,35 @@
+import React, { Component } from "react";
+import { MarkerComponentProps } from "./MapResultsView";
+import { css } from "emotion";
+
+class DefaultMarker<Marker> extends Component<MarkerComponentProps<Marker>> {
+  render() {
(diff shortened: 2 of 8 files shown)


zlatanpham:
Hey @benhinchley, I tried to write some doc but not sure why `yarn docs:dev` fails to run. So please refer to the following snippet as an example of usage:

```tsx
import React, { Component } from 'react';
import {
  MapResultsView,
  MarkerRemote,
  MarkerComponentProps,
} from '@sajari/sdk-react';

interface CustomMarkerProps {
  picture_url: string;
  space_type: string;
  bed_label: string;
  name: string;
  price: string;
}

interface MarkerProps extends MarkerRemote<CustomMarkerProps> {}

interface AppState {
  data: {
    lat: number;
    lng: number;
    id: string;
    markers: MarkerRemote<CustomMarkerProps>[];
  };
}

class MapResultsViewExtend extends MapResultsView<CustomMarkerProps> {}

class CustomMarker extends React.Component<MarkerComponentProps<CustomMarkerProps>> {
  render() {
    const {
      isActive,
      isSelected,
      name,
      picture_url,
      getMarkerProps,
      ...rest
    } = this.props;
    return (
      <div className="custom-marker" {...getMarkerProps()}>
        {
          // JSX go here
        }
      </div>
    );
  }
}

class App extends Component<{}, AppState> {
  render() {
    return (
      <MapResultsViewExtend
        options={{ styles: customMapStyle, zoomControl: false }}
        GoogleAPIMapKey="YOUR_API_KEY"
        MarkerComponent={CustomMarker}
        data={this.state.data}
      />
    );
  }
}
```

The generic component accepts to pass a `React` component to the `MarkerComponent` prop so it gives the freedom to build UI for markers.

I'm not sure it's necessary to allow passing callbacks like `onDragEnd` or `onZoomChange` or controlling the maker state from outside components or not. So it would be better if you can give a thought about that.

PR #103, opened 2019-02-11
Enhance custom styles

### LINKED ISSUE #93: Update to emotion 10
Emotion 10 has come out, which has a series of useful features added. 

We should also figure out a way to turn off all styling created by us (leaving all the static CSS classnames) for when people want a clean slate when styling

zlatanpham:
resolve #93


zlatanpham:
@benhinchley I think the process should be similar as just wrapping `withGetStyles` around the components which requires custom `styles` prop.

Issue #102, opened 2019-01-20
Create expandable search interface for Sajari website and website search integrations


alexluscombe:
Develop an updated version of the search interface on the Sajari website using the React SDK. New designs can be found here:

Search box and 5 overlay results: https://app.zeplin.io/project/5ba096204799c73235ed771c/screen/5c0db33282be7db71db52e74 (new design on the right)

Full results page:
https://app.zeplin.io/project/5ba096204799c73235ed771c/screen/5c0db33182be7db71db52d7c

**Behavior:**
- Initial animation is similar to the version currently on the Sajari homepage i.e. field expands and placeholder text is removed when field is clicked
- Search field should use `Instant search and typeahead`
- Top 5 results should be displayed by default alongside filters (see design)
- Link at bottom of results `Show all xxx results` clicks through to full paginated results page where 10 results are shown on each page by default

zlatanpham:
Hi @alexluscombe, can you take a look at [demo](https://sajari-new-search-interface.netlify.com/) for the new interface and give your feedback?  and please note that:
- There is no responsiveness yet. If it's in need, please inform me to do the part.
- The custom checkboxes customized from the `checkbox` sdk do not play nicely with the keyboard control. I'll try to fix that later.
- The design of results page includes 2 search bars. I'm not sure if it makes sense to have 2 search bars sharing the same state on a page, thus, I removed the search input from the header of the [demo](https://sajari-new-search-interface.netlify.com/). Please inform me if it still required to keep both of them.

alexluscombe:
Hi @zlatanpham thanks for the demo. It's looking good but we have a few changes.

• Add a 68px margin on the bottom of the overlay results box, so it doesn’t touch the end of (short) pages. Reproduce by dragging the page to make it smaller to touch the bottom of the results box.
• Need to correct how hyperlinks work. Please find the original hyperlink styling here. - https://jsfiddle.net/vanejung/faLqvwnc/
• Need to correct the textfield shadow behaviours on hover and focus state on the results page.  Should only show shadow when field has been clicked, should be in non-active state when the results page is initially rendered as per design here - https://app.zeplin.io/project/5ba096204799c73235ed771c/screen/5c512cdb31131d309fbf37e3
• Adjusted design on results page so that filters are on left hand side. Updated design here. - https://app.zeplin.io/project/5ba096204799c73235ed771c/screen/5c512cdb31131d309fbf37e3
• Removing search from header on results is a good idea, thanks.
[...]

PR #104, opened 2019-02-13
Search: wrap summary data to Search component

zlatanpham:
**WHAT**:
I am using the `Search` component to build the interface for #102 but the UI requires data like `totalResults`, `q.original`, etc and I have to use `Consumer` to get these data.

**WHY**:
It's cumbersome to write nested render-prop components since we can do an update to use just `Search` only.

**HOW**:
Get all the needed data from `pipeline` methods like `getResults ` or `getResponse` and inject those to `summary` prop and pass to render prop function.

### DIFF
--- a/src/components/Search/Search.tsx
+++ b/src/components/Search/Search.tsx
@@ -24,28 +24,14 @@ export interface PipelineProps {
   summary: SummaryInterface | undefined;
 }
 
-export interface QueryValuesInterface {
-  filter: string;
-  q: string;
-  page: string;
-}
-
-export interface ValuesInterface {
-  q?: string;
-  "q.original"?: string;
-  "q.suggestions"?: string;
-}
-
-export interface ResponseInterface {
-  reads: number;
-  totalResults: number;
-  time: number;
+export interface PipelineResponseInterface {
+  [k: string]: string | number;
 }
 
 export interface SummaryInterface {
-  queryValues: QueryValuesInterface;
-  values: ValuesInterface;
-  response: ResponseInterface;
+  queryValues: PipelineResponseInterface;
+  values: PipelineResponseInterface;
+  response: PipelineResponseInterface;
 }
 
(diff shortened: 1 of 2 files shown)


src/components/Search/Search.tsx
```diff
@@ -105,6 +112,18 @@ export class Search extends React.PureComponent<SearchProps<any>, {}> {
               pipelines.instant.response.getResults()) ||
               []
           );
+
+          const response =
+            pipelines.search.response || pipelines.instant.response;
+          let summary = {};
+          if (response) {
+            summary = {
+              ...mapToObject(response.getQueryValues()),
```
benhinchley: The following two lines have the potential to override each other, `q` in Query values will get overridden by `q` in responseValues if autocomplete is running. It might be worth having a `queryValues`, and `responseValues` key in the `SummaryInterface`.
```js
...mapToObject(response.getQueryValues()),
...mapToObject(response.getValues()),
```

src/components/Search/Search.tsx
```diff
@@ -21,11 +21,31 @@ export interface PipelineProps {
...
+export interface ValuesInterface {
+  q?: string;
+  "q.original"?: string;
+  "q.suggestions"?: string;
+}
+
+export interface ResponseInterface {
+  reads: number;
+  totalResults: number;
+  time: number;
```
zlatanpham: @benhinchley I've updated the code as your suggestion and also explicitly made the keys of `queryValues`,..  visible rather than using `[key: string]: number | string`. Not sure it's ok to you because I guess those properties would be permanent.

benhinchley: Hey @zlatanpham the values for a query, can be anything as a user can customise the variables needed for their pipeline. So the correct type for these objects is actually `{ [k:string]: string | number }`

zlatanpham: updated!

PR #108, opened 2019-05-02
Input: fix hard to click drag to select text

zlatanpham:
This PR aims to resolve https://github.com/sajari/website-search-integration/issues/11

WHAT:
It's hard for users to click + drag to select the text on the search box.

WHY:
The input was wrapped by `React-Autosize-Input` component in order to position . This implementation made the width of the input not span over its div container so users will actually click + drag on the div container rather than the input.

HOW:
- Remove the `React-Autosize-Input` component and force the width of the input to span over its parent box. 
- Keep the typeahead word without trimming, place it on the left-most of the search box and vertically center it. Finally, hide the substring of the actual input by `opacity:0`.

Here is the demo - https://codesandbox.io/s/8804qlkr02.

### DIFF
--- a/src/components/Input/Typeahead.tsx
+++ b/src/components/Input/Typeahead.tsx
@@ -12,20 +12,33 @@ interface TypeaheadProps {
 
 const typeaheadStyles: CSSObject = {
   color: "#bebebe",
-  display: "inline",
   fontSize: "1em",
-  marginLeft: -2,
-  position: "relative"
+  position: "absolute",
+  width: "100%",
+  overflow: "hiddden",
+  left: 0,
+  top: "50%",
+  transform: "translateY(-50%)",
+  pointerEvents: "none",
+  span: {
+    opacity: 0
+  }
 };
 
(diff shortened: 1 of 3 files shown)


src/components/Input/Typeahead.tsx
```diff
@@ -12,20 +12,33 @@ interface TypeaheadProps {
 
 const typeaheadStyles: CSSObject = {
   color: "#bebebe",
-  display: "inline",
   fontSize: "1em",
-  marginLeft: -2,
-  position: "relative"
+  position: "absolute",
```
benhinchley: I would also put `zIndex: -1,` on this so it sits behind the cursor, but just check that it still looks okay in firefox, etc after that change, I have noticed some odd things can happen in the past.

zlatanpham: I don't think `zIndex: -1` is necessary since `pointer-events: none` does the trick to block any interaction with the node.

PR #110, opened 2019-08-21
fix select Input returns No Results prompt

zlatanpham:
**WHAT**:
when the user presses Tab and comes on the search box, a "No Results" prompt appears for the Voiceover utility

**WHY**:
~Because the downshift `getA11yStatusMessage` function will return "No results." if the list of items is empty - https://github.com/downshift-js/downshift#geta11ystatusmessage~
When input is focused, it will trigger `openmenu` which will call `getA11yStatusMessage`. Initially, there is no search results so the status message will be `no results` which will be called out by screen reader.

**HOW**:
~Passing a function to return empty string to disable the behavior (have tested and screen reader still works properly on suggestion dropdown)~
Checking to call `openmenu` only when at least one result is available.

### DIFF
--- a/src/components/Input/Input.tsx
+++ b/src/components/Input/Input.tsx
@@ -92,7 +92,6 @@ export class Input extends React.PureComponent<InputProps, InputState> {
   public render() {
     return (
       <Search
-        getA11yStatusMessage={() => ""}
         stateReducer={this.stateReducer}
         key={`sj-downshift-${this.props.defaultValue}`}
         defaultInputValue={this.props.defaultValue}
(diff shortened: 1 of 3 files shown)


src/components/context/pipeline/Provider.tsx
```diff
@@ -301,7 +301,7 @@ function debounce<F extends Procedure>(
     isImmediate: false
   }
 ): F {
-  let timeoutId: NodeJS.Timeout | undefined;
```
zlatanpham: this was causing build fail

src/components/Input/Input.tsx
```diff
@@ -92,6 +92,7 @@ export class Input extends React.PureComponent<InputProps, InputState> {
   public render() {
     return (
       <Search
+        getA11yStatusMessage={() => ""}
```
benhinchley: So this just disables the `No results` message?

zlatanpham: @benhinchley understanding the issue. I'm looking for another way to resolve the problem.

PR #116, opened 2020-02-24
Add 'All' option for checkbox

tuanddd:
## What this PR does

Add an `all` props to `<Checkbox />` component to allow for check-all option.

### DIFF
--- a/src/components/Checkbox/Checkbox.tsx
+++ b/src/components/Checkbox/Checkbox.tsx
@@ -4,6 +4,7 @@ import { Container, HiddenInput, NativeInput } from "./styled";
 
 export interface CheckboxProps {
   name: string;
+  all?: boolean;
   CheckboxRenderer?: React.ComponentType<CheckboxRendererProps>;
 }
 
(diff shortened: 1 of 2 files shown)


src/components/Checkbox/Checkbox.tsx
```diff
@@ -13,16 +14,25 @@ export interface CheckboxRendererProps {
...
+            : selected.includes(name);
           return (
             <Container
               isSelected={isSelected}
-              onClick={this.onClick(name, isSelected, set)}
+              onClick={
+                all
+                  ? () => {
+                      Object.keys(options).forEach(k => set(k, false));
+                      return;
```
zlatanpham: redundant `return`

tuanddd: Fixed.

tuanddd:
> This looks good, though just wondering what the need to add the `all` prop to the `Checkbox` component is?
> 
> The checkbox acts as a radio button by default, so selecting all should reset to the `All` defined in the filter.

Hi @benhinchley, the props is needed for visual purpose, because there are cases that mutlple checkboxes are allowed, so selecting the `All` checkbox should deselect the others and vice versa (selecting at least one of the others will deselect the `All` checkbox). Hope this makes sense to you.

PR #120, opened 2020-03-10
Add onChange props to Paginator component

tuanddd:
### What this PR does
Currently, the `Paginator` component provides no method to listen for page change to the outside components so the UI cannot update accordingly, this will add a `onChange` props that is invoked whenever page is change whether it's jumping to specific page, navigate backward or forward.

### DIFF
--- a/src/components/Paginator/Paginator.tsx
+++ b/src/components/Paginator/Paginator.tsx
@@ -10,6 +10,12 @@ import { PaginatorContainer } from "./Container";
 import { LeftChevron, RightChevron } from "./icons";
 import { Container, PageButton, PageNumber } from "./styled";
 
+export interface PaginatorState {
+  pageNumber: number;
+  totalPages: number;
+  triggeredBy: "backward" | "forward" | "jump" | "firstRender";
+}
+
 export interface PaginatorProps {
   windowSize?: number;
   className?: string;
@@ -22,7 +28,12 @@ export interface PaginatorProps {
   PreviousButtonRenderer?: React.ComponentType<PageButtonProps>;
   NextButtonRenderer?: React.ComponentType<PageButtonProps>;
   PageNumberRenderer?: React.ComponentType<PageNumberProps>;
+  onChange?: (state: PaginatorState) => void;
 }
+
+const Wrapper = (props: any) => {
+  return props.children();
+};
 
 export function Paginator(props: PaginatorProps) {
   const {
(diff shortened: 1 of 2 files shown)


src/components/Paginator/Paginator.tsx
```diff
@@ -22,61 +28,90 @@ export interface PaginatorProps {
...
+              React.useEffect(() => {
+                // tslint:disable-next-line: no-unused-expression
+                onChange &&
+                  typeof onChange === "function" &&
+                  onChange({
+                    pageNumber: page,
+                    totalPages,
+                    triggeredBy: "firstRender"
+                  });
+              }, []);
```
zlatanpham: I'm not keen on the idea of putting a hook inside an anonymous function. Also, what is the purpose of this? why do we need to trigger `onChange` at the first render?

src/components/Paginator/Paginator.tsx
```diff
@@ -147,23 +182,55 @@ function DefaultPageNumber(props: PageNumberProps) {
...
+) {
   return function closure(event: React.MouseEvent<HTMLButtonElement>) {
     if (page === 1) {
       return;
     }
-
     setPage(paginate, page - 1)(event);
+    // tslint:disable-next-line: no-unused-expression
+    onChange &&
+      typeof onChange === "function" &&
```
zlatanpham: Defensive coding. Checking onChange not being undefined is enough.

src/components/Paginator/Paginator.tsx
```diff
@@ -147,23 +182,55 @@ function DefaultPageNumber(props: PageNumberProps) {
...
+) {
   return function closure(event: React.MouseEvent<HTMLButtonElement>) {
     if (page === totalPages) {
       return;
     }
 
     setPage(paginate, page + 1)(event);
+    // tslint:disable-next-line: no-unused-expression
+    onChange &&
+      typeof onChange === "function" &&
```
zlatanpham: same here

src/components/Paginator/Paginator.tsx
tuanddd: Done

src/components/Paginator/Paginator.tsx
tuanddd: Done

src/components/Paginator/Paginator.tsx
tuanddd: The idea is to allow for external component to be able to view the paginator state on first render, but it is indeed a little bit weird to fire `onChange` on first render, I'm removing this.

benhinchley:
I'm not 100% sure of the use case for this change, when the page is changed the values object gets updated with the latest page number, which could be used to solve this case I believe.

I may be missing context on this change.

tuanddd:
> I'm not 100% sure of the use case for this change, when the page is changed the values object gets updated with the latest page number, which could be used to solve this case I believe.
> 
> I may be missing context on this change.

Your method might work but last time I checked the `<Consumer />` component doesn't have access to the `values` object so maybe instead of this add `values` to the context?

PR #124, opened 2020-04-14
rangeSlider component

zlatanpham:
This PR aims to create RangeSlider which will be used to narrow down search results, for example, based on pricing.

- [x] `rangeFilter`
- [x] `rangeAggregrateFilter`
- [x] `RangeSlider` component
- [x] ability to work with `rangeAggregrateFilter`
- [x] being customized in terms of styles, extra UI for tip, etc

### DIFF
--- a/src/components/RangeSlider/RangeSliderUI.tsx
+++ b/src/components/RangeSlider/RangeSliderUI.tsx
@@ -22,7 +22,11 @@ export const RangeSliderUI = ({
     return (
       <Handle {...hanleModifiedProps}>
         <ValueTip>
-          {renderTip ? renderTip(handleProps.value) : handleProps.value}
+          {renderTip ? (
+            <span>{renderTip(handleProps.value)}</span>
+          ) : (
+            <span>{handleProps.value}</span>
+          )}
         </ValueTip>
       </Handle>
     );
--- a/src/components/RangeSlider/RangeSliderUI.tsx
+++ b/src/components/RangeSlider/RangeSliderUI.tsx
@@ -1,18 +1,33 @@
 import classnames from "classnames";
 import { Handle, RangeProps } from "rc-slider";
 import React from "react";
-import { StyledRanger, ValueTip } from "./styled";
+import { CSSObject } from "styled-components";
+import { StyledRange, ValueTip } from "./styled";
 
-export interface RangeSliderUIProps extends RangeProps {
-  renderTip?: (value: number) => JSX.Element;
-  className?: string;
+export interface RangeCustomStyleObject {
+  handle?: CSSObject;
+  container?: CSSObject;
+  rail?: CSSObject;
+  track?: CSSObject;
 }
+
+export interface RangerSliderCustomProps {
+  className?: string;
+  styles?: RangeCustomStyleObject;
+  renderTip?: (value: number) => JSX.Element;
+  allowCross?: boolean;
+  pushable?: boolean | number;
+  step?: number;
+}
+
+export type RangeSliderUIProps = RangeProps & RangerSliderCustomProps;
 
 export const RangeSliderUI = ({
   renderTip,
   className,
   ...props
 }: RangeSliderUIProps) => {
+  // Assign type `any` for handleProps due to missing definition from lib
   const handle = (handleProps: any) => {
     const hanleModifiedProps = {
       ...handleProps,
@@ -33,7 +48,7 @@ export const RangeSliderUI = ({
   };
 
   return (
-    <StyledRanger
+    <StyledRange
       {...props}
       className={classnames("sj-range-slider", className)}
       defaultValue={props.value}
--- a/src/components/RangeSlider/RangeSliderUI.tsx
+++ b/src/components/RangeSlider/RangeSliderUI.tsx
@@ -1,7 +1,7 @@
+import { CSSObject } from "@emotion/core";
 import classnames from "classnames";
 import { Handle, RangeProps } from "rc-slider";
 import React from "react";
-import { CSSObject } from "styled-components";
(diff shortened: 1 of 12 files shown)


sampotts:
@zlatanpham I'll review once the styling is done perhaps? (cc @jschumacher)

zlatanpham:
> I'll review once the styling is done perhaps?
Yes, there are a couple of small issues with the AggregrateFilter as well. I'll request your review later once getting those done.

src/components/RangeSlider/RangeSliderUI.tsx
```diff
@@ -1,7 +1,7 @@
+import { CSSObject } from "@emotion/core";
 import classnames from "classnames";
 import { Handle, RangeProps } from "rc-slider";
 import React from "react";
-import { CSSObject } from "styled-components";
```
zlatanpham: Ahh, we use `emotion` for styling but I used the autocomplete function from VSCode to inject the type and it wrongly picked the one from `styled-components`. I've pushed the code to fix that.

zlatanpham:
> ## Handles can overtake each other
> I think we shouldn't allow the handles to overtake each other. So when the left handle can only butt up against the right handle rather than going past it.
> 
> ## Formatted numbers (nice to have)
> It'd be nice if the numbers were humanized with thousands seperators but this is a nice to have if it's a big bit of work. It's easily possible using [`Intl.NumberFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NumberFormat) but if we have to support super old IE then it'd need polyfilling. I wasn't sure if we had something for it already?

@sampotts have you had the chance to look at the docs - https://deploy-preview-124--sajari-sdk-react.netlify.app/components/rangeSlider? We can custom the overtake behavior via `pushable` and `allowCross` prop and custom the format of the number by passing a custom `jsx` through `renderTip`.

sampotts:
> @sampotts have you had the chance to look at the docs - https://deploy-preview-124--sajari-sdk-react.netlify.app/components/rangeSlider? We can custom the overtake behavior via `pushable` and `allowCross` prop and custom the format of the number by passing a custom `jsx` through `renderTip`.

Ah, I didn't know we had examples in the docs. That's fine then 👍

zlatanpham:
> # Proposed design
> react-sdk-slider
> 
> This design takes into account the above issues. The left handle is the `:focus`/`:active` state. The fill color should use their theme color and we can default it to whatever we use for the other components. If there's no color then grab me on Slack and we can work something out. There is 40px padding/margin above the track and 15px below. It doesn't have to match exactly but close would be good to give the slider some breathing space.

- Yes, you're right about the spacing.
- The current implement does handle `:focus` state but yes, it doesn't work properly like you said. One thing I just notice the focus/active state will disappear once you use left/right arrow to control the handle. Is that the thing you want me to fix?

Anyway, will update the style to follow the new design 😃

sampotts:
>     * The current implement does handle `:focus` state but yes, it doesn't work properly like you said. One thing I just notice the focus/active state will disappear once you use left/right arrow to control the handle. Is that the thing you want me to fix?

Yes please, that'd be good if it kept the focus style while using keyboard input.

sampotts:
Looks good to me 👍 Nice one. I guess we just need @benhinchley's seal of approval and we're good?

src/components/RangeSlider/rangeSlider.mdx
```diff
@@ -0,0 +1,126 @@
...
+---
+
+import { PropsTable } from "docz";
+import { RangeFilter } from "../../controllers";
+import { RangeSlider } from "./RangeSlider.tsx";
+import { Playground } from "docz";
+import { EVENT_SELECTION_UPDATED } from "../../events";
+
+# RangeSlider
+
```
benhinchley, line 14: The `RangeSlider` component provides a slider which constructs the filter for the search request. It is customizable through the use of props which allows you to apply different styles, set different step value, enable or disable crossover, restrict a margin, and render a prefix.

benhinchley: ^ Sorry @zlatanpham somehow lost the message I wrote on the review comment, if you could just add the above paragraph in this change is good to go.

zlatanpham: I've added the message.

sampotts:
>     * I needed to add `theme: any;` in the `ContainerProps` interface in `src/components/Tabs/Tab/styled.ts` for the build to work.

Hey @zlatanpham, did this get sorted? ☝️

zlatanpham:
> > ```
> > * I needed to add `theme: any;` in the `ContainerProps` interface in `src/components/Tabs/Tab/styled.ts` for the build to work.
> > ```
> 
> 
> Hey @zlatanpham, did this get sorted? ☝️

@sampotts I mentioned the reason here - https://github.com/sajari/sajari-sdk-react/pull/124#discussion_r411090692 plus the commit to fix it - https://github.com/sajari/sajari-sdk-react/pull/124/commits/f895cc7ce9b1b02cfa1fdd8284f601c50d342e99. Are you still experiencing the issue?

sampotts:
> @sampotts I mentioned the reason here - [#124 (comment)](https://github.com/sajari/sajari-sdk-react/pull/124#discussion_r411090692) plus the commit to fix it - [f895cc7](https://github.com/sajari/sajari-sdk-react/commit/f895cc7ce9b1b02cfa1fdd8284f601c50d342e99). Are you still experiencing the issue?

Yeah I'm still seeing it on latest. A customer also raised it. I've opened a PR here with a fix and also an unrelated docs update: 
https://github.com/sajari/sajari-sdk-react/pull/125

PR #131, opened 2020-07-23
rename `name` to `id` and use that for filter instead of title

### LINKED ISSUE #107: Tabs map to filters using displayname instead of name

tuanddd:
Fixes #107 
- [x] Rename `name` key to `id` to make more sense and use it to for filter id instead of title (displayname)


tuanddd:
Got it 👍

PR #142, opened 2020-09-20
setup styling framework

zlatanpham:
- [x] allow using tailwind classes via `twin.macro`.
- [x] toggle `!important` option via `ThemeProvider` prop.
- [x] option for theming.


src/components/Button/index.tsx
```diff
@@ -0,0 +1 @@
+export * from "./Button";
```
sampotts, line 1: Why do we use this pattern of having two files for one component? Should we just export the button as default in this file?

tailwind.config.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
```
sampotts, line 1: We can use es exports here can't we?

src/components/Button/index.tsx
zlatanpham: I'll make the update. Since old components follow the structure so it happened to me to just keep the same pattern for the sample component. Personally, for building apps, I prefer using the pattern for IDE file searching but for building a component library I agree we should export components as default for the sake of tree-shaking. Seeing many libraries such as `lodash`, `material-ui` or `antd` as they set up and export their components in a particular structure so we can use [babel-plugin-import](https://github.com/ant-design/babel-plugin-import) to ignore unwanted modules.

tailwind.config.js
zlatanpham: Yes, we can. Initially, I created the file by tailwind cli and had no touch to it.

zlatanpham: Hmm, changing it to `export default {` will break the `docz` build. Maybe for the future but for now, can we still keep the CommonJS syntax?

sampotts: Ah yeah for sure, this is minor. Leave it as CJS. I think the plan down the road is to move to Next.js + MDX like we have for sajari-ui so this may become a non-issue.

src/components/Box/index.tsx
```diff
@@ -1 +1,23 @@
...
+import React from "react";
+import { PropsWithAs } from "../../types/component-as";
+import { forwardRefWithAs } from "../../utils/forwardRefWithAs";
+
+interface BoxProps {
+  children?: React.ReactNode;
+}
+
+const BoxComponent = (
+  props: PropsWithAs<BoxProps, "div">,
```
sampotts, line 10: Should we store "div" in a constant? Minor but perhaps `const defaultElement = "div"` might look a little tidier.

zlatanpham: @sampotts `div` here is for typing so should the change be like below?
```tsx
type DefaultElement = "div";

const BoxComponent = (
  props: PropsWithAs<BoxProps, DefaultElement>,
  ref: React.Ref<HTMLDivElement>
) => {
  const { as: Type = "div", children, ...rest } = props;
  return (
    <Type ref={ref} {...rest}>
      {children}
    </Type>
  );
};

const Box = forwardRefWithAs<BoxProps, DefaultElement>(BoxComponent);
```

sampotts: Looks good to me! 👍

sampotts (approved):

PR #143, opened 2020-09-21
Add utility for handling `as` prop

tuanddd:


tuanddd:
@sampotts Please wait til I include the new `Box` component in this change, too.

PR #150, opened 2020-09-25
Add Swatch component

tuanddd:
- [x] Add predefined color set
- [x] Allow user to add their own color
- [x] Ensure accessibility


src/components/Swatch/color.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+const StyledLabel = styled.label<{
+  color: string;
+  bg: string;
+  border: string;
+  focus: boolean;
+}>`
+  ${tw`flex items-center justify-center rounded-full w-8 h-8 border border-solid`};
+  box-shadow: ${({ focus }) =>
+    focus ? "0 0 0 3px rgba(66, 153, 225, 0.5)" : "none"};
```
sampotts, line 30: Is there not a tailwind equivalent of this shadow?

src/components/Swatch/color.tsx
```diff
@@ -0,0 +1,152 @@
...
+  color: ${({ color }) => color};
+  background-color: ${({ bg }) => bg};
+  border-color: ${({ border }) => border};
+`;
+
+const colors: ColorProps[] = [
+  { id: "White", bg: "white", color: "#718096", border: "#E2E8F0" },
```
sampotts, line 37: Is there no way to get these color values from the tailwind config?

src/components/Swatch/icons.tsx
```diff
@@ -0,0 +1,3 @@
+import React from 'react';
+
+export const Check = (props: any) => <svg viewBox="0 0 16 16" width="16" height="16" {...props}><path d="M14.3 2.3L5 11.6 1.7 8.3c-.4-.4-1-.4-1.4 0s-.4 1 0 1.4l4 4c.2.2.4.3.7.3s.5-.1.7-.3l10-10c.4-.4.4-1 0-1.4s-1-.4-1.4 0z"/></svg>
```
sampotts, line 3: As with the other change, perhaps we have all icons in one spot?

tuanddd: For some reasons I can't `export { ReactComponent as Check } from './check.svg'`, tried adding a declaration file but then met with a new bug. I have to temporarily import the file directly rather than from the `index.tsx`.

src/components/Swatch/color.tsx
tuanddd: Done

src/components/Swatch/color.tsx
tuanddd: Done

src/components/Swatch/icons.tsx
sampotts: Ah maybe try importing first, then exporting? 

```javascript 
import { ReactComponent as IconName } from './icon-name.svg';

export { IconName }
```

sampotts (approved):

PR #152, opened 2020-09-29
V3: Add Rating component

tuanddd:
- [x] Half rating
- [x] Default icon
- [x] Add support for custom icon
- [x] Ensure a11y


src/components/Rating/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+        ) : null}
+        {Array.from(Array(remainder)).map((k, i) => (
+          <RatingItem
+            index={i}
+            count={max}
+            character={character}
+            active={false}
+          />
+        ))}
+        <span css={tw`sr-only`}>{`${value} stars`}</span>
```
zlatanpham, line 64: Should also care about singular here?

src/components/Rating/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+        {children ? children : null}
```
zlatanpham, line 65: Can it just be
```suggestion
        {children}
```

src/components/Rating/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+    return (
+      <StyledBox flipped={flipped} ref={ref} role="img" aria-label={label}>
+        {Array.from(Array(Math.floor(value))).map((k, i) => (
+          <RatingItem
+            index={i}
+            count={max}
+            character={character}
+            active={true}
+          />
```
zlatanpham, line 44: Look like a missing `key` prop here

src/components/Rating/RatingItem.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const characterNode =
+    typeof character === "function" ? character() : character;
+
+  return (
+    <Box
+      css={
+        active && !half
+          ? tw`relative m-0 p-0 inline-block text-orange-400`
+          : tw`relative m-0 p-0 inline-block text-gray-300`
+      }
```
zlatanpham, line 40: I wonder if we can do something like this:
```suggestion
      css={[tw`relative m-0 p-0 inline-block`, active && !half ? tw`text-orange-400` : tw`text-gray-300`]}
```
then it can save you few bytes of text once the code gets transformed :)

src/components/Rating/RatingItem.tsx
```diff
@@ -0,0 +1,51 @@
+/** @jsx jsx */
+import { jsx } from "@emotion/core";
+import React from "react";
+import tw, { styled, theme } from "twin.macro";
+import Box from "../Box";
```
zlatanpham, line 5: In case we don't need `as` prop, I wonder if we should just use a `div` rather than using `Box`. It'll just save you one unnecessary function call.

src/components/Rating/types.ts
```diff
@@ -0,0 +1,44 @@
...
+  /** Its self pos */
+  index: number;
+  /** Total count */
+  count: number;
+}
+
+export enum ItemType {
+  FILLED = 1,
+  HALF_FILLED,
+  EMPTY
```
sampotts, line 41: Is this the general style we use? I'd suggest we stick to `CamelCase` here?

tuanddd: Yeah somehow I keep tripping myself at this.

PR #153, opened 2020-09-29
V3: Add Label component

zlatanpham:


src/components/Label/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+import React from 'react';
+import { __DEV__ } from '../../utils/assersion';
+import useLabelStyles from './styles';
+import { LabelProps } from './types';
+
+
+const Label = React.forwardRef((props: LabelProps, ref?: React.Ref<HTMLLabelElement>) => {
+  const styles = useLabelStyles(props)
+
+  return <label {...props} ref={ref}  css={styles} />;
```
sampotts, line 12: There's an extra space here before `css=` but I think we just need to set up prettier (and possibly ESLint) on this repo. Will discuss in the meeting.

sampotts (approved):

PR #158, opened 2020-10-01
V3: Add Text component

tuanddd:
A clone from the component library

### DIFF
--- a/src/utils/style-props.ts
+++ b/src/utils/style-props.ts
@@ -1,0 +1,29 @@
+import tw, { TwStyle } from "twin.macro";
+import { isNumber } from "../utils/assertion";
+
+export function mapTruncateValue(truncate: boolean | 2 | 3 | 4 | 5): TwStyle {
+  if (truncate === true) {
+    return tw`truncate`;
+  }
+
+  if (isNumber(truncate)) {
+    switch (truncate) {
+      case 2:
+        return tw`truncate-2-lines`;
+
+      case 3:
+        return tw`truncate-3-lines`;
+
+      case 4:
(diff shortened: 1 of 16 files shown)


src/utils/assersion.ts
```diff
@@ -2,6 +2,10 @@ export function isArray<T>(value: any): value is T[] {
   return Array.isArray(value);
 }
 
+export function isNumber(value: any): value is number {
```
sampotts, line 5: Is `value is number` correct? Or should it be `boolean`?

src/utils/style-props.ts
```diff
@@ -0,0 +1,29 @@
+import tw, { TwStyle } from "twin.macro";
+import { isNumber } from "../utils/assersion";
```
jonathaningram, line 2: assersion->assert (or assertion)

src/utils/assersion.ts
tuanddd: I think it is correct? It is taken from `sajari-ui`.

src/utils/style-props.ts
tuanddd: Can I put this in another PR? Some components are referencing this file and changing it will polute this PR.

src/utils/assersion.ts
sampotts: Ah, you're right! 👍

src/utils/style-props.ts
jonathaningram: I'm easy, whatever works.

sampotts (approved):

PR #159, opened 2020-10-05
V3: Add RangeInput component

tuanddd:
- [x] Ensure a11y
- [x] Add input
- [x] Add support for custom input component


src/components/RangeSlider/RangeSliderUI.tsx
```diff
@@ -34,8 +34,10 @@ export const RangeSliderUI = ({
       dragging: handleProps.dragging.toString()
     };
 
+    const { index } = handleProps;
+
     return (
-      <Handle {...hanleModifiedProps}>
+      <Handle key={`handle-${index}`} {...hanleModifiedProps}>
```
tuanddd, line 40: Not necessarily relevant to this PR but this fixes the warning.

src/components/RangeInput/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+        <RangeSliderUI
+          min={min}
+          max={max}
+          onChange={handleRangeChange}
+          value={range}
+          {...rest}
+        />
+        <div css={tw`flex flex-col sm:flex-row items-center justify-between`}>
+          {leftInput}
+          <hr
```
sampotts, line 115: Not sure `<hr>` is the right element here? It would be ready as a horizontal rule in screen readers I think?

tuanddd: What do you suggest we put here?

sampotts: How does it render?

tuanddd: Pretty much like this, the dash in the middle 
Screen Shot 2020-10-07 at 1 19 50 PM

sampotts: Ah in that case, it should be an `&ndash;`. We'll likely want to tweak that design before we go live but it's fine for now.

tuanddd:
@sampotts I haven't convert this yet because it depends on a controller (`RangeFilter`), so maybe we'll have the controllers merged first then we can get back to this?

sampotts:
> @sampotts I haven't convert this yet because it depends on a controller (`RangeFilter`), so maybe we'll have the controllers merged first then we can get back to this?

Sure, sounds good.

packages/components/src/RangeInput/styled.ts
```diff
@@ -0,0 +1,62 @@
+import { styled } from 'twin.macro';
```
sampotts, line 1: I'll look at the styling of this later on and see if we can use the tailwind styling here.

sampotts (approved):

sampotts:
@tuanddd I think this just needs the conflicts sorting 👍

tuanddd:
> @tuanddd I think this just needs the conflicts sorting 👍

Oh sorry will do it right now ☹️

tuanddd:
@sampotts Once the tests passed it's good to go!

PR #165, opened 2020-10-08
V3: Setup monorepo

zlatanpham:
- [x] setup monorepo including `@sajari/react-sdk`, `@sajari/react-hooks`, `@sajari/react-components`, `sajari-sdk-docs` and `sajari-react-sdk-utils` for common utilities.
- [x] build with `tsdx`.
- [x] setup testing with `react-testing-library`.
- [x] setup docs and deploy it to Netlify.
- [x] move already merged components to `@sajari/react-components` package.


.eslintrc.js
```diff
@@ -0,0 +1,36 @@
+module.exports = {
```
sampotts, line 1: Can this not just be `.eslintrc` ?

docs/.eslintrc
```diff
@@ -0,0 +1,51 @@
+{
+  // comment out because it gave a warning
+  // "extends": ["@sampotts/eslint-config/react-ts"],
```
sampotts, line 3: Maybe just remove this for now. Ideally, we should have a `@sajari/eslint-config` config that we're all happy with.

docs/components/snippets.ts
```diff
@@ -0,0 +1,54 @@
+export default {
```
sampotts, line 1: Do we need this file?

docs/components/make-data.ts
```diff
@@ -0,0 +1,18 @@
+import { loremIpsum } from 'lorem-ipsum';
```
sampotts, line 1: Do we need this file?

docs/components/utils.ts
```diff
@@ -0,0 +1,16 @@
+/* eslint-disable no-param-reassign, import/prefer-default-export */
```
sampotts, line 1: Not sure if we need this file since how we handle icons in this lib is different.

packages/components/src/Image/index.tsx
```diff
@@ -0,0 +1,99 @@
+/** @jsx jsx */
+import { jsx } from '@emotion/core';
+import React, { useEffect, useRef, useState } from 'react';
+import tw from 'twin.macro';
+import Box from '../Box';
+
+import { __DEV__ } from '../utils/assersion';
```
sampotts, line 7: I think this file name needs to be `assertion` which @tuanddd updated in another PR.

.eslintrc.js
zlatanpham: Will update that.

docs/components/snippets.ts
zlatanpham: 👍  I will scan through the source to remove unused files.

sampotts (approved):

PR #173, opened 2020-10-19
V3: add Context wrapper

zlatanpham:
- [x] add SearchContextProvider built on top of `@sajari/sdk-js@next`.
- [x] add field mapping for a custom collection.
- [x] add `useSearchContext` hook as an example of how we will next build hooks to consume the data from the upper context. The detail of usage is provided in [./example](https://github.com/sajari/sajari-sdk-react/compare/v3...zlatanpham:v3-add-context-wrapper-and-field-mapping?expand=1#diff-efd419d19cf8794f3925f3d2af3f3919cca43bd7bc4e412d33dfed88f0f33d41).

Note that other than the new changes listed above, most of the others like `Pipeline`, `Values`, and `Tracking` stuff are directly copied from `v1` with a minor tweak to make them work with the new APIs from `@sajari/sdk-js@next`. There are a couple of TODO notes in the PR related to the APIs change that might be better being fixed in another PR.


packages/hooks/src/SearchContextProvider/index.tsx
```diff
@@ -0,0 +1,309 @@
...
+  name: 'PipelineContext',
+});
+
+interface State {
+  response: Response | null;
+  query: string;
+  completion: string;
+  suggestions: string[];
+  config: Config;
+}
```
sampotts, line 147: As above

packages/hooks/src/utils/mapResultFields.ts
```diff
@@ -0,0 +1,26 @@
+import { Result } from '@sajari/sdk-js';
+
+import { Fields } from '../SearchContextProvider';
+
+export function mapFields<T = Record<string, string | string[]>>(values: T, fields: Fields = {}) {
+  const mappingFields = { ...values };
+
+  // eslint-disable-next-line no-restricted-syntax
+  for (const key of Object.keys(fields)) {
```
sampotts, line 9: Why not `Object.entries(fields).forEach(([key, value]) => {...}` ?

sampotts (approved):

PR #177, opened 2020-10-21
V3: feat/add usePagination hook

tuanddd:
## Should merge #179  first


packages/hooks/package.json
```diff
@@ -17,7 +17,7 @@
     "dist"
   ],
   "dependencies": {
-    "@sajari/sdk-js": "next",
+    "@sajari/sdk-js": "^2.0.0-alpha.12",
```
sampotts, line 20: `next` would resolve the same version

sampotts (approved):

PR #184, opened 2020-10-26
V3: fix: pageSize not reactive

tuanddd:
## Why does this PR exist?
Initially `usePageSize` composed `usePagination` - particularly `pageSize` variable - but since that value in `usePagination` is derived from the `response` which is only updated between search request so it doesn't make sense as page size is something user can update client-side and as a result changes to `pageSize` from `usePageSize` cannot be picked up by React.

## What does this PR do?
- [x] Change the `pageSize` to be derived from the `values` itself.

### DIFF
--- a/packages/hooks/src/usePageSize/index.ts
+++ b/packages/hooks/src/usePageSize/index.ts
@@ -1,6 +1,5 @@
 import React from 'react';
 import { useContext } from '../SearchContextProvider';
-import usePagination from '../usePagination';
 import { UsePageSizeResult } from './types';
 
 function usePageSize(): UsePageSizeResult {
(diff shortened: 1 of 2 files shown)


sampotts (approved):

packages/hooks/src/usePageSize/index.ts
```diff
@@ -19,8 +17,10 @@ function usePageSize(): UsePageSizeResult {
     [values],
   );
 
+  const pageSize = parseInt(values.get()[resultsPerPageParam], 10);
+
   return {
-    pageSize,
+    pageSize: Number.isNaN(pageSize) ? 10 : pageSize,
```
sampotts, line 23: Just a note but I think we should have a "defaults" file somewhere for stuff like this. I'd say the default here should be 15 anyway.

PR #185, opened 2020-10-26
V3: feat: add PageSize composition

tuanddd:


packages/search-ui/src/PageSize/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+            {s}
+          </option>
+        ))}
+      </Select>
+    </div>
+  );
+};
+
+if (__DEV__) {
+  PageSize.displayName = 'PageSizeComposition';
```
sampotts, line 32: I think we should just call this `PageSize`. Can be updated later though of course.

sampotts (approved):

PR #180, opened 2020-10-22
v3: add useQuery and useSearch hook

zlatanpham:
- [x] add useQuery
- [x] add useSearch
- [x] add docs

PR #187, opened 2020-10-26
V3: add results composition

zlatanpham:
This change depends on https://github.com/sajari/sajari-sdk-react/pull/180


sampotts:
Once the conflicts om this one and #180 are sorted then we can merge 👍

PR #195, opened 2020-10-28
V3: feat: allow range input no filter

tuanddd:
## Should merge #194 first.


docs/pages/components/rangeinput.mdx
```diff
@@ -29,10 +29,77 @@ function Example() {
...
+### Usage without Filter controller
+
+In case you don't want to depend on `RangeFilter`, you can otp-out of that by passing `limit`, `value`, `onChange` to the `RangeInput`.
+
+```jsx
+function Example() {
+  const [range, setRange] = React.useState([200, 100]);
+  return (
+    <div className="flex flex-col">
+      <RangeInput limit={[0, 500]} value={range} onChange={setRange} />
```
sampotts, line 87: Should `limit` really be split into `min` and `max` instead so we're similar to the native element's API?

tuanddd: Isn't the `limit` props already include the `min` and `max`? 🤔
The syntax is `[min, max]` but if you wanna split that up that's fine also fine 🙂

docs/pages/components/rangeinput.mdx
```diff
@@ -29,10 +29,78 @@ function Example() {
...
+}
+```
+
+## Props
+
+| Name         | Type                                     | Default    | Description                                                                                       |
+| ------------ | ---------------------------------------- | ---------- | ------------------------------------------------------------------------------------------------- |
+| `min`        | `number`                                 | `0`        | The min range, used when you want to use the component without RangeFilter                        |
+| `max`        | `number`                                 | `100`      | The max range, used when you want to use the component without RangeFilter                        |
+| `value`      | `[number, number]`                       | `[25, 50]` | The current value of the range input, used when you want to use the component without RangeFilter |
```
sampotts, line 102: Sorry, I should have mentioned this earlier, is there a way to make this a single value if you just want one slider?

docs/pages/components/rangeinput.mdx
```diff
@@ -29,10 +29,78 @@ function Example() {
...
+| `onChange`   | `(value: [number, number]) => void`      | `() => {}` | The callback when the value changes, used when you want to use the component without RangeFilter  |
```
sampotts, line 103: Same as above I guess for the value

packages/components/src/RangeInput/defaults.ts
```diff
@@ -0,0 +1,7 @@
+import { Range } from '@sajari/react-hooks';
+
+export const defaultParams = {
+  onChange: () => {},
+  value: [25, 50] as Range,
+  limit: [0, 100] as Range,
```
sampotts, line 6: I guess this needs updating?

PR #197, opened 2020-10-28
V3: chore: polish

tuanddd:
# Should merge #194 first.

## What does this PR do?
- [x] Make some changes to `Result` (a11y) and its doc.
- [x] Make some changes to `<Results />` composition (styling and logic).
- [x] Replace bits of `<Result />` + loop with `<Results />` composition.
- [x] Implement `searchOnChange` on `PageSize` and `Sorting` composition.

### DIFF
--- a/packages/search-ui/src/Results/index.tsx
+++ b/packages/search-ui/src/Results/index.tsx
@@ -1,6 +1,6 @@
 /** @jsx jsx */
 import { jsx } from '@emotion/core';
-import { useSearch } from '@sajari/react-hooks';
+import { useSearchContext } from '@sajari/react-hooks';
 import React from 'react';
 
 import Result from '../Result';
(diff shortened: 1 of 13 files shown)


packages/search-ui/src/Results/index.tsx
```diff
@@ -1,14 +1,14 @@
...
-import { useSearch } from '@sajari/react-hooks';
+import { useSearchContext } from '@sajari/react-hooks';
 import React from 'react';
 
 import Result from '../Result';
 import useResultsStyles from './styles';
 import { ResultsProps, ResultValues } from './types';
 
 const Results = (props: ResultsProps) => {
-  const { results } = useSearch<ResultValues>();
```
tuanddd, line 11: Using `useSearch` will trigger two search requests wherever `<Results />` is used because internally `useSearch` also send a search request when the query value changes.

sampotts: Should useSearch only execute a search if the values actually change? I've had this before where if the `useEffect` dependency is an object or array, there's no deep inspection done to determine if values changed. The only way to get around it is to use `JSON.stringify(object)`.

sampotts (approved):

PR #200, opened 2020-11-02
V3: refactor: rewrite useSearch hook

tuanddd:
## What does this PR do?
Rewrite the `useSearch` to be able to search when the request object changes
- [x] Use `useSearch` instead of `useSearchContext` whenever possible
- [x] Add `loading`, `error` as return value of `useSearch`
- [x] Allow for query override via `useSearch` parameter
- [x] Allow for `Pipeline`, `Values`, `fields` override

It should be noted that when this PR is merged all hooks change will trigger search request

### DIFF
--- a/packages/hooks/src/useSearch/index.ts
+++ b/packages/hooks/src/useSearch/index.ts
@@ -1,113 +1,131 @@
 /* eslint-disable @typescript-eslint/no-shadow */
-import { useCallback, useEffect, useState } from 'react';
+import { useCallback, useEffect, useRef, useState } from 'react';
 
 import { useContext } from '../SearchContextProvider';
-import { Pipeline, Response, Values } from '../SearchContextProvider/controllers';
-import { EVENT_RESPONSE_UPDATED } from '../SearchContextProvider/events';
-import { FieldDictionary } from '../SearchContextProvider/types';
+import { defaultConfig } from '../SearchContextProvider/config';
+import { Response } from '../SearchContextProvider/controllers';
+import { EVENT_RESPONSE_UPDATED, EVENT_VALUES_UPDATED } from '../SearchContextProvider/events';
+import useQuery from '../useQuery';
 import debounce from '../utils/debounce';
 import mapResultFields from '../utils/mapResultFields';
-import mapToObject from '../utils/mapToObject';
+import { UseSearchCustomConfig, UseSearchParams, UseSearchResult } from './types';
 
-// Returns results for the current query
-function useSearchDefaultQuery<T>() {
-  const {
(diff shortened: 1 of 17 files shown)


packages/hooks/src/useSearch/index.ts
```diff
@@ -1,113 +1,131 @@
...
-  } = useContext();
-
-  const { time: latency, totalResults } = mapToObject<{ time?: number; totalResults?: number }>(
-    response?.getResponse() as Map<string, any> | undefined,
-  );
-  const reqResults = response?.getResults();
+import { UseSearchCustomConfig, UseSearchParams, UseSearchResult } from './types';
 
+function useCustomSearch({ pipeline, values, fields = {} }: UseSearchCustomConfig): UseSearchResult {
+  const [searching, setSearching] = useState(false);
```
sampotts, line 14: Is there a reason we don't just use `loading`?

packages/hooks/src/utils/debounce.ts
```diff
@@ -1,31 +1,29 @@
...
+
+export default function useDebounce<F extends (...args: any[]) => void>(
   func: F,
   waitMilliseconds = 50,
   options = {
     isImmediate: false,
   },
 ): F {
-  let timeoutId: number | undefined;
+  const timeoutIdRef = useRef<number | undefined>();
```
sampotts, line 10: `const timeoutIdRef = useRef<ReturnType<typeof setTimeout>();`

packages/hooks/src/utils/debounce.ts
```diff
@@ -1,31 +1,29 @@
...
 
   // eslint-disable-next-line func-names
   return function (this: any, ...args: any[]) {
+    window.clearTimeout(timeoutIdRef.current);
```
sampotts, line 14: I think we can ditch the window reference?

packages/hooks/src/utils/debounce.ts
```diff
@@ -1,31 +1,29 @@
...
-    const shouldCallNow = options.isImmediate && timeoutId === undefined;
-
-    if (timeoutId !== undefined) {
-      clearTimeout(timeoutId);
-    }
+    const shouldCallNow = options.isImmediate && timeoutIdRef.current === undefined;
 
-    // @ts-ignore
-    timeoutId = setTimeout(doLater, waitMilliseconds);
+    timeoutIdRef.current = window.setTimeout(doLater, waitMilliseconds);
```
sampotts, line 26: As above, `window.setTimeout` -> `setTimeout`

sampotts (approved):

PR #204, opened 2020-11-09
V3: add list filter composition

zlatanpham:
The PR depends on https://github.com/sajari/sajari-sdk-react/pull/201


sampotts:
I had originally thought the pattern for the Filter compositions would be with a type prop `<Filter type="list" />`, do you think that would work? Possibly save some double up of logic?

zlatanpham:
> I had originally thought the pattern for the Filter compositions would be with a type prop `<Filter type="list" />`, do you think that would work? Possibly save some double up of logic?

Will make the update. But the `list` and `color` filter do not share the same options so probably the Filter interface will look like:

```ts
interface ListProps{
 type: 'list',
 ....
}

interface ColorProps{
  type: 'color',
  ...
}

type FilterProps = ListProps | ColorProps;
```

zlatanpham:
Closing in favor of https://github.com/sajari/sajari-sdk-react/pull/206

PR #206, opened 2020-11-10
V3: add Filter composition

zlatanpham:
Added `Filter` component for `search-ui` package. Currently,  it supports `list` and `color` type.

### DIFF
--- a/docs/components/CodeBlock.tsx
+++ b/docs/components/CodeBlock.tsx
@@ -2,7 +2,7 @@ import { mdx } from '@mdx-js/react';
 
 import * as ReactComponents from '@sajari/react-components';
 import * as ReactHooks from '@sajari/react-hooks';
-import * as ReactSDK from '@sajari/react-search-ui';
+import * as ReactSearchUI from '@sajari/react-search-ui';
 import SearchProvider from './SearchProvider';
 
 import {
@@ -19,6 +19,7 @@ import { useState } from 'react';
 import { LiveEditor, LiveError, LivePreview, LiveProvider } from 'react-live';
 import * as ReactTable from 'react-table';
 import resolveConfig from 'tailwindcss/resolveConfig';
+import { useRouter } from 'next/router';
 
 const { theme } = resolveConfig(tailwindConfig);
 
@@ -89,6 +90,15 @@ const CodeBlock = (props: CodeBlockProps) => {
   const [editorCode, setEditorCode] = useState(children.trim());
   const language = (className && (className.replace(/language-/, '') as Language)) || 'jsx';
   const { onCopy, hasCopied } = useClipboard(editorCode);
+  const { pathname } = useRouter();
+
+  // We have to do some manual work here to avoid naming conflict between components
+  let extra = {};
+  if (pathname === '/search-ui/filter') {
+    extra = { Filter: ReactSearchUI.Filter, FilterOptions: ReactHooks.Filter };
+  } else if (pathname === '/hooks/usefilter') {
+    extra = { Filter: ReactHooks.Filter };
+  }
 
   const liveProviderProps = {
     theme: codeThemes.dark,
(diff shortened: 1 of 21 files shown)


docs/components/CodeBlock.tsx
```diff
@@ -89,6 +90,15 @@ const CodeBlock = (props: CodeBlockProps) => {
...
   const { onCopy, hasCopied } = useClipboard(editorCode);
+  const { pathname } = useRouter();
+
+  // We have to do some manual work here to avoid naming conflict between components
+  let extra = {};
+  if (pathname === '/search-ui/filter') {
+    extra = { Filter: ReactSearchUI.Filter, FilterOptions: ReactHooks.Filter };
+  } else if (pathname === '/hooks/usefilter') {
+    extra = { Filter: ReactHooks.Filter };
+  }
```
zlatanpham, line 101: I'm still not able to figure out a better way to handle the naming conflicts right now rather than "physically" getting it working like ^.

packages/hooks/src/index.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable import/export */
```
sampotts, line 1: What was this complaining about?

packages/search-ui/src/Filter/ColorFilter.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+  if (filtered.length === 0) {
+    return null;
+  }
+
+  return (
+    <Box title={title} showReset={selected.length > 0} onReset={reset}>
+      <Swatch checkedColors={selected} onChange={setSelected}>
+        {filtered.map((color) => {
+          const Component = Swatch.Color[color];
```
sampotts, line 23: Will this allow for custom color values?

packages/search-ui/src/Filter/ListFilter.tsx
```diff
@@ -0,0 +1,102 @@
+/* eslint-disable react/jsx-curly-brace-presence */
```
sampotts, line 1: What's this one for?

packages/search-ui/src/Filter/utils.ts
```diff
@@ -0,0 +1,18 @@
+import { FilterItem } from '@sajari/react-hooks';
+
+// Move selected items to top
+export function sortList(list: FilterItem[], selected: string[]) {
```
sampotts, line 4: Do we not need to sort the filter items by the count value too? I wrote something to do this for the shopify embed: https://github.com/sampotts/shopify-embed/blob/master/src/utils/object.ts

packages/hooks/src/index.ts
zlatanpham: I managed to remove the line but not for the below. Basically, it complains `Parsing error: Cannot read property 'map' of undefined`. Googling for a while but I still cannot find a solution for that.

packages/search-ui/src/Filter/ColorFilter.tsx
zlatanpham: No. The colors are derived from those pre-defined in the `Swatch` component and the component itself does not allow custom color values. Is this something we should take note of for later implementation?

packages/search-ui/src/Filter/ListFilter.tsx
zlatanpham: Removed. Seems I wanted to quickly test something but forgot to remove the line.

packages/search-ui/src/Filter/utils.ts
zlatanpham: Updated.

docs/components/CodeBlock.tsx
sampotts: Yeh I'm not sure there's any other way to do it...

packages/hooks/src/index.ts
sampotts: Ah, I had this before. I'll see if I can fix it.

sampotts (approved):

PR #214, opened 2020-11-12
V3: fix: search on empty string with custom filter

tuanddd:
## What this PR does
- [x] Allow search on empty query
- [x] Add custom filter `_id != ""`


packages/hooks/src/utils/assertion.ts
```diff
@@ -0,0 +1,34 @@
+/* eslint-disable no-underscore-dangle */
```
sampotts, line 1: We really need to move a load of utils to the utils package to avoid replication. Should we add a TODO or should I just add a ticket?

sampotts (approved):

PR #217, opened 2020-11-12
V3: improve combobox UX

zlatanpham:
This PR aims to retain the functionality we had in v2

### DIFF
--- a/docs/pages/components/combobox.mdx
+++ b/docs/pages/components/combobox.mdx
@@ -69,6 +69,38 @@ function Example() {
 }
 ```
 
+### Results mode
+
+```jsx
+function Example() {
+  const items = [
+    {
+      title: 'HEXBUG - Battle Spider Battle Ground - Multi',
+      image: 'https://cdn.sajari.net/demos/bestbuy/images/5235810_sb.jpg',
+      url: 'https://www.bestbuy.com/site//5235810.p',
+      description:
+        'Experience a revolutionary fight with this Hexbug Battle Spider Battle Ground. Two remote control spiders fight to the end with infrared sensors that have 1.5-meter fire ranges. Each spider has a 360-degree rotating head and LED eyes that let you steer it toward its opponent. This Hexbug Battle Spider Battle Ground comes with 15 walls for a fun, challenging match.',
+    },
+    {
+      title: 'Beats by Dr. Dre - Beats Studio Wireless Over-Ear Headphones - Unity Edition Red White Blue',
+      image: 'https://cdn.sajari.net/demos/bestbuy/images/5655629_rb.jpg',
+      url: 'https://www.bestbuy.com/site//5655629.p,
+      description:
+        'Beats Studio Wireless Unity Edition combines iconic Beats sound and design with country pride elements. Each headphone incorporates country-inspired graphic elements with a unifying black band for a signature Beats look.',
+    },
+    {
+      title: 'Epson - 664 Ink Bottle - Cyan',
+      image: 'https://cdn.sajari.net/demos/bestbuy/images/4274202_sb.jpg',
+      url: 'https://www.bestbuy.com/site//4274202.p',
+      description:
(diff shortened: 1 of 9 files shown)


docs/pages/components/combobox.mdx
```diff
@@ -58,6 +58,42 @@ function Example() {
...
+    <Combobox
+      mode="suggestions"
+      label="Query"
+      name="query"
+      value={query}
+      placeholder="Search"
+      items={items}
+      onChange={setValue}
+      itemToString={(item) => item.title}
+      itemRender={() => <div>{item.title}</div>}
```
sampotts, line 91: I know this is just the `suggetions` example but I think we'd want to handle this a little smarter for the users. Perhaps by default, `suggestion` items just render a simple text string (no need for the superfluous div) and the `results` option would render the `Results` or `Result` component if `itemRender`.

docs/pages/components/combobox.mdx
```diff
@@ -58,6 +58,42 @@ function Example() {
...
+function Example() {
+  const [query, setQuery] = React.useState('');
+  const [value, setValue] = React.useState('');
+  const items = [
+    { title: 'apple iphone' },
+    { title: 'apple imac' },
+    { title: 'apple watch' },
+    { title: 'apple mac pro' },
+    { title: 'apple macbook pro' },
+  ];
```
sampotts, line 73: It might be good if we used real data in the example?

packages/components/src/Combobox/components/DropdownResult/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+  return (
+    <Box
+      as="li"
+      {...getItemProps({
+        index,
+        item: value,
+      })}
+      key={`${value}_${index}`}
+    >
+      <a href={value.url} css={styles.item}>
```
sampotts, line 25: Should we not be using the token here?

packages/components/src/Combobox/components/DropdownResult/index.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { jsx } from '@emotion/core';
+import { Heading, Image, Text } from '@sajari/react-components';
+import tw from 'twin.macro';
+
+import Box from '../../../Box';
+import { useComboboxContext } from '../../context';
+import { useDropdownItemStyles } from './styles';
+import { DropdownResultProps } from './types';
+
+const DropdownResult = (props: DropdownResultProps) => {
```
sampotts, line 11: Just wondering why we can't use the `Result` component here? Is it because we don't export it from the sajari-ui package?

zlatanpham: No, it's not the main reason. It's because we are handling click tracking in the Result component and it will cause the Combobox to not purely functional. We can make the Result component be representational and export it from the `components` package but since we had the Input composition, where I think we should put to use the `Result` component instead. My thought for the interface is:

```jsx
// Combobox.tsx
interface ComboboxProps<T>{
   mode: 'suggestions' | ...,
   itemRender?: (props: ({item: T, ...})) => React.ReactNode,
   // ...rest
}

// Input.tsx
import {Combobox, ComboboxProps} from '@sajari/react-components';
import {Result} from '../Results';

interface InputProps extends Omit<ComboboxProps , 'itemRender' | 'items'>{
   // ...rest
}

const InputProps = (props: InputProps) => {
   const {results} = useSearch();

   return <Combobox {...props} items={results} itemRender={({item}) => <Result {...item} />} />;
}
```
My initial work was to do them all in one PR but finally decided to separate the work into 2 PRs. This one only aims to ensure we don't miss the accessibility handling we had in v2 and the other is to complete the Input composition which I'm already progressing on.

sampotts: Sounds logical to me. Thanks for the explanation 👍

sampotts (approved):

PR #222, opened 2020-11-15
fix: performance

tuanddd:
## What this PR does
- [x] Change `useMemo` to `options` to prevent UI blocking/slow down
- [x] Remove ref on `Result` as it's unnecessary, replaced with memo

### DIFF
--- a/packages/search-ui/src/Results/Result/index.tsx
+++ b/packages/search-ui/src/Results/Result/index.tsx
@@ -16,81 +16,90 @@ const Heading = styled(Text)`
   ${tw`font-medium text-gray-900`}
 `;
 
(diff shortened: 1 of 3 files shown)


sampotts (approved):

packages/search-ui/src/Results/Result/index.tsx
```diff
@@ -16,81 +16,90 @@ const Heading = styled(Text)`
...
-  const { handleResultClicked } = useTracking();
-  let clickToken;
-  if (token && 'click' in token) {
-    clickToken = token.click;
-  }
-
-  const resultClicked = React.useCallback(() => {
-    if (url) {
-      handleResultClicked(url);
+const Result = React.memo(
```
sampotts, line 19: Why do we need this to be a memo?

PR #223, opened 2020-11-15
fix: search function in useSearch

tuanddd:
## What this PR does
- [x] Modify the check to be more specific

### DIFF
--- a/packages/hooks/src/useSearch/index.ts
+++ b/packages/hooks/src/useSearch/index.ts
@@ -67,7 +67,7 @@ function useNormalSearch(queryOverride: string = ''): UseSearchResult {
 
   const search = useCallback(
     (q?: string) => {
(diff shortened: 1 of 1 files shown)


packages/hooks/src/useSearch/index.ts
```diff
@@ -67,7 +67,7 @@ function useNormalSearch(queryOverride: string = ''): UseSearchResult {
 
   const search = useCallback(
     (q?: string) => {
-      if (q) {
+      if (q !== undefined) {
```
tuanddd, line 70: We only want to search for the current query if `q` is `undefined` so the following behavior is correct
```js
search() // resend search request for current query
```
vs
```js
search("") // search for ""
```

sampotts (approved):

PR #225, opened 2020-11-16
V3: chore: tidy up search logic

tuanddd:
## Why this PR exists
Currently, when we select any filter option, 2 requests are sent, 1 is from the `"selection-updated"` event handler that directly invokes the search function, and 2 is from `useSearch`'s logic of running effect whenever the value object changes.

This case isn't ideal and there is a potential of a loop occurring, the search function sets the values object -> which are picked up by `useEffect` -> which again runs the search function

## What this PR does
- [x] Remove the effect logic in `useSearch` entirely, so now all is handed to the search function
- [x] As a result, all hooks and composition that depends on that (`usePageSize`/`PageSize`, ...etc) must have an additional call to search function to preserve the reactivity
- [x] Some section of the docs are also updated to instant search instead of having to press enter - which isn't intuitive given the example's UI

## Result
Before the change, `useFilter` section of the docs when rendered will fire more than 3 requests which doesn't add up with the number of examples, and choosing any filter fires 2 requests as said above
After the change, only 3 requests are fired on mount, and only 1 request for each action after (input, filter)

@sampotts if you can please do a sanity check again in your local before merging to prevent any bugs, cheers :)

### DIFF
--- a/docs/pages/hooks/usepagesize.mdx
+++ b/docs/pages/hooks/usepagesize.mdx
@@ -30,20 +30,14 @@ function Example() {
   const SearchPlayground = React.memo(() => {
     const { search, results } = useSearch();
     const { pageSize, setPageSize } = usePageSize();
-    const [query, setQuery] = React.useState('');
+    const { query } = useQuery();
 
     return (
       <div className="flex flex-col space-y-4">
         <div className="flex space-x-4">
-          <form
-            className="w-3/5"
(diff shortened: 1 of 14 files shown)


sampotts:
@tuanddd Ran this PR locally and checked the Filters and they seemed to work fine for me. They only made one request when filtering 👍

sampotts (approved):

PR #227, opened 2020-11-17
v3: feat: updated props for Filter component

sampotts:
This change adds several props to the `Filter` component: 

- `sort` which allows choosing how the filter is sorted (`'count'` (by the numeric count), `'alpha'` (using the label) or `'none'` (random order))
- `sortAscending` - default `true` if sort is not `'count'`
- `pinSelected` - pin selected options to the top of the list - default `true` 

It changes `multi` and `count` to true in the `Filter` Class by default since I think this would be the most common use-case. 

It also renames `repeated` to `array` to match the latest API naming changes.

### DIFF
--- a/docs/pages/hooks/usefilter.mdx
+++ b/docs/pages/hooks/usefilter.mdx
@@ -252,17 +252,13 @@ function Example() {
 
   const colorFilter = new Filter({
     name: 'color',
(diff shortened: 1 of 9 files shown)


zlatanpham (approved):
LGTM 👍

PR #230, opened 2020-11-18
V3: fix wrong import in combobox causing docs failed to build

zlatanpham:


packages/components/src/Combobox/components/DropdownResult/index.tsx
```diff
@@ -19,7 +21,6 @@ const DropdownResult = (props: DropdownResultProps) => {
         index,
         item: value,
       })}
-      key={`${value}_${index}`}
```
zlatanpham, line 22: Remove `key` prop since it doesn't look like to do anything necessary here.

sampotts: I added it to try and fix a warning but the warning still showed anyway 🤷‍♂️

sampotts (approved):

PR #231, opened 2020-11-18
V3: minor fixes

tuanddd:
## What this PR does
- [x] Modify the Travis CI script so that it will fail on error ([because by default it won't and will try to run all commands](https://docs.travis-ci.com/user/job-lifecycle/#customizing-the-build-phase))
- [x] Fix some filter tests due to the change of `count` and `multi` being true by default
- [x] Fix `sortItems` and `pinItems` tests
- [x] Fix some lint warning, sorry I had my extension disabled without knowing it
- [x] Fix a import error

### DIFF
--- a/packages/search-ui/src/Filter/utils.ts
+++ b/packages/search-ui/src/Filter/utils.ts
@@ -8,7 +8,12 @@ export function sortItems(list: Record<string, any>[], prop?: string, asc: boole
   return [...list].sort((a, b) => {
     const l = prop ? a[prop] : a;
     const r = prop ? b[prop] : b;
-    return asc ? l - r : r - l;
+    if (l === r) return 0;
+    if (asc) {
+      if (l > r) return 1;
+      return l > r ? 1 : -1;
+    }
+    return l > r ? -1 : 1;
   });
 }
 
--- a/packages/search-ui/src/Filter/utils.ts
+++ b/packages/search-ui/src/Filter/utils.ts
@@ -4,16 +4,21 @@
  * @param {Boolean} asc - Ascending order?
  * @param {String} prop - Property of child object to sort on
  */
+const collator = new Intl.Collator(typeof window !== 'undefined' ? window.navigator.language : 'en-US', {
+  numeric: true,
+});
+
 export function sortItems(list: Record<string, any>[], prop?: string, asc: boolean = true) {
   return [...list].sort((a, b) => {
     const l = prop ? a[prop] : a;
     const r = prop ? b[prop] : b;
-    if (l === r) return 0;
+
     if (asc) {
-      if (l > r) return 1;
-      return l > r ? 1 : -1;
+      return collator.compare(String(l), String(r));
     }
-    return l > r ? -1 : 1;
+
+    // Invert the result;
+    return collator.compare(String(r), String(l));
   });
 }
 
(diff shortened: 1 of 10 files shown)


sampotts (approved):

sampotts (approved):

packages/search-ui/src/Filter/utils.ts
```diff
@@ -8,7 +8,12 @@ export function sortItems(list: Record<string, any>[], prop?: string, asc: boole
   return [...list].sort((a, b) => {
     const l = prop ? a[prop] : a;
     const r = prop ? b[prop] : b;
-    return asc ? l - r : r - l;
+    if (l === r) return 0;
+    if (asc) {
+      if (l > r) return 1;
+      return l > r ? 1 : -1;
```
sampotts, line 14: The first line here is redundant, isn't it? It does the same as the first part of the ternary.

packages/search-ui/src/Filter/utils.ts
```diff
@@ -8,7 +8,12 @@ export function sortItems(list: Record<string, any>[], prop?: string, asc: boole
   return [...list].sort((a, b) => {
     const l = prop ? a[prop] : a;
     const r = prop ? b[prop] : b;
-    return asc ? l - r : r - l;
```
sampotts, line 11: What was wrong with this logic? AFAIK the sort function just needs to return a negative number, positive number or `0` to determine the sorting.

packages/search-ui/src/Filter/utils.ts
tuanddd: Ah yeah my bad.

packages/search-ui/src/Filter/utils.ts
tuanddd: Subtracting only works for numbers, if the value is string then you got `NaN` and the array is not sorted.

sampotts: Ah yeah, good point.

tuanddd: I've changed it to be locale dependent using `Intl.Collator`, this is the best option between `localeCompare` and normal comparison ([best speed and has locale support](https://stackoverflow.com/a/52369951))

packages/search-ui/src/Filter/utils.ts
```diff
@@ -4,16 +4,31 @@
...
+  // eslint-disable-next-line no-console
+  console.error(error);
+}
+
 export function sortItems(list: Record<string, any>[], prop?: string, asc: boolean = true) {
   return [...list].sort((a, b) => {
     const l = prop ? a[prop] : a;
     const r = prop ? b[prop] : b;
-    if (l === r) return 0;
+    // Fallback to the normal comparison operator if Collator is not supported;
```
sampotts, line 19: Seems a bit complex? Why not something like this:

```javascript
if (typeof l === 'string') {
  const comparison = l.localeCompare(r);
  return asc ? compare : compare * -1; 
}

return asc ? l - r : r - l;
```

tuanddd: Ah those are not my latest change (forgot to commit them) but anyway let's go with yours :)

sampotts (approved):

PR #233, opened 2020-11-19
V3: fix AspectRatio not preventing overflow

tuanddd:
## What this PR does
- [x] Add `overflow-hidden` to the `AspectRatio` to prevent overflowing

(Please ignore the branch name now, it's irrelevant)

### DIFF
--- a/packages/components/src/AspectRatio/styles.ts
+++ b/packages/components/src/AspectRatio/styles.ts
@@ -5,7 +5,7 @@ import { AspectRatioProps } from './types';
 
 export function useAspectRatioStyles(props: AspectRatioProps) {
   const { ratio } = props;
-  const styles: (string | TwStyle)[] = [];
+  const styles: (string | TwStyle)[] = [tw`overflow-hidden`];
 
   if (!ratio) {
     return css(styles);
--- a/docs/pages/components/aspectratio.mdx
+++ b/docs/pages/components/aspectratio.mdx
@@ -30,11 +30,11 @@ to the link of the video.
 
 ### Embed Image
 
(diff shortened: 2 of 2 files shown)


sampotts:
What's the reason behind this change? When would we nest `AspectRatio`? It just seems riskier to change this.

packages/components/src/AspectRatio/styles.ts
```diff
@@ -5,7 +5,7 @@ import { AspectRatioProps } from './types';
 
 export function useAspectRatioStyles(props: AspectRatioProps) {
   const { ratio } = props;
-  const styles: (string | TwStyle)[] = [];
+  const styles: (string | TwStyle)[] = [tw`overflow-hidden`];
```
sampotts, line 8: Could we not just add this when needed? e.g. `<AspectRatio css={tw`overflow-hidden`}>`. I think it should be an opt-in change or it might break things like Tooltips, etc.

tuanddd: We could but if you head over to https://sajari-react-sdk-v3.netlify.app/components/aspectratio and refresh a few times you can see that the Embed Image example has a bug.

sampotts: Maybe just change the image example to use either `<Image src="https://source.unsplash.com/random" aspectRatio={4 / 3} objectFit="cover" alt="Photo" />` or native ``. Using `overflow: hidden` would just break the appearance I think.

tuanddd: Cool okay I'll close this one.

PR #235, opened 2020-11-19
V3: implement input composition

zlatanpham:
This PR aims to implement `Input` composition by integrating `Combobox` component with `useSearch` hook. Some changes has been made in `Combobox` to have it support a render prop function as well as adding a couple of missing `a11y` functionality from the last update.


packages/search-ui/src/Input/index.tsx
```diff
@@ -1,13 +1,121 @@
...
+                      const link = clickToken || url.toString();
+                      if (link) {
+                        handleResultClicked(link);
+                        window.location.href = url.toString();
+                      }
+                    },
+                  })}
+                  css={styles.item}
+                >
+                  <ResultView
```
zlatanpham, line 103: The `suggestion` mode uses the `Result` component as `Dropdown` item but I think we might want to write another component for that because it looks too detailed and also includes unnecessary `a` tags.

sampotts: Maybe we can just use the same `DropdownResult`?

packages/search-ui/src/Results/Result/index.tsx
```diff
@@ -7,31 +7,81 @@ import React from 'react';
...
+            <div css={tw`ml-6`}>
+              <Text>{formatNumber(Number(price), currencyCode)}</Text>
+            </div>
+          )}
+        </div>
+
+        {(category || typeof rating === 'number') && appearance === 'list' && (
+          <div css={tw`flex items-baseline mt-1`}>
+            {category && <Text css={tw`mr-3 text-xs text-gray-400`}>{category}</Text>}
+            {typeof rating === 'number' && <Rating value={rating} max={ratingMax} />}
```
sampotts, line 61: We should use `isNumber` from assertion here but I think we can revert this anyway as mentioned earlier in the PR.

sampotts (approved):

PR #238, opened 2020-11-20
V3: fix Input standard mode broken when typing in

zlatanpham:
The change aims to fix the issue as when typing in the Input composition in `standard` mode will cause the error `TypeError: n.startsWith is not a function`.

This PR also includes the update to make sure the examples from docs compatible with the previous changes from Input composition.

### DIFF
--- a/packages/search-ui/src/Results/index.tsx
+++ b/packages/search-ui/src/Results/index.tsx
@@ -9,10 +9,10 @@ import { ResultsProps, ResultValues } from './types';
 const Results = (props: ResultsProps) => {
   const { results } = useSearchContext<ResultValues>();
   const styles = useResultsStyles(props);
-  const { appearance = 'list', ...rest } = props;
+  const { appearance = 'list', className, ...rest } = props;
 
   return (
-    <div css={[styles.container]}>
+    <div css={[styles.container]} className={className}>
       {results?.map(({ values: { category, description, image, _id, price, rating, title, url }, token }) => (
         <Result
           token={token}
--- a/packages/search-ui/src/Results/index.tsx
+++ b/packages/search-ui/src/Results/index.tsx
@@ -1,23 +1,25 @@
 /** @jsx jsx */
 import { jsx } from '@emotion/core';
+import { Box } from '@sajari/react-components';
 import { useSearchContext } from '@sajari/react-hooks';
+import { isEmpty } from '@sajari/react-sdk-utils';
 
 import Result from './components/Result';
 import useResultsStyles from './styles';
 import { ResultsProps, ResultValues } from './types';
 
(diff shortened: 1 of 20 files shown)


sampotts:
image

In the [Netlify preview](https://deploy-preview-238--sajari-react-sdk-v3.netlify.app/search-ui/input) I get the above after typing into the input for suggestions. Also, does the `Results` component not accept a `style` and/or `className` prop? If not we should add that before launch as it looks a little odd with the extra margin when there are no results.

sampotts:
Another bug I found is if I head to the [filter docs page](https://deploy-preview-238--sajari-react-sdk-v3.netlify.app/search-ui/filter) and perform a search such as `'f'` it performs a search, then if I delete the f so the query is empty, the results don't revert back to those for an empty search. Here's a GIF to show what I mean... 

CleanShot 2020-11-23 at 10 47 47

packages/search-ui/src/Results/index.tsx
```diff
@@ -7,13 +7,17 @@ import useResultsStyles from './styles';
...
+  const { appearance = 'list', className, ...rest } = props;
+
+  if (results.length === 0) {
+    return null;
+  }
 
   return (
-    <div css={[styles.container]}>
-      {results?.map(({ values: { category, description, image, _id, price, rating, title, url }, token }) => (
+    <div css={[styles.container]} className={className}>
```
sampotts, line 19: Should we make this a `Box` and just allow `className` on the `Box`? I'm thinking that might be cleaner since we'll want `className` elsewhere too.

packages/search-ui/src/Results/types.ts
```diff
@@ -6,6 +6,7 @@ export interface ResultsProps {
   imageAspectRatio?: ImageProps['aspectRatio'];
   imageObjectFit?: ImageProps['objectFit'];
   currencyCode?: string;
+  className?: string;
```
sampotts, line 9: As above, if we use `Box` then we can just `extend BoxProps`?

packages/search-ui/src/Results/index.tsx
```diff
@@ -7,13 +7,17 @@ import useResultsStyles from './styles';
 import { ResultsProps, ResultValues } from './types';
 
 const Results = (props: ResultsProps) => {
-  const { results } = useSearchContext<ResultValues>();
+  const { results = [] } = useSearchContext<ResultValues>();
   const styles = useResultsStyles(props);
-  const { appearance = 'list', ...rest } = props;
+  const { appearance = 'list', className, ...rest } = props;
+
+  if (results.length === 0) {
```
sampotts, line 14: Should we use `isEmpty` instead?

sampotts (approved):

PR #240, opened 2020-11-23
V3: add mode to input

tuanddd:
## What this PR does
- [x] Add an addtional mode `"instant"` to `<Input />` since we will also allow user to be able to search on submit, hence `"standard"`

### DIFF
--- a/docs/pages/search-ui/input.mdx
+++ b/docs/pages/search-ui/input.mdx
@@ -48,6 +48,31 @@ import { Input } from 'sajari/react-search-ui';
 </>
 ```
 
+## Standard
+
+By default `<Input />` has an additional mode `"instant"` and it is also the default value (search on change), if you want to disable it, simple change mode to `"standard"`:
+
+```jsx
+function Example() {
+  const [value, setValue] = React.useState('');
+
+  return (
+    <form
+      onSubmit={(event) => {
+        event.preventDefault();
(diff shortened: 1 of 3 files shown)


docs/pages/search-ui/input.mdx
```diff
@@ -48,6 +48,23 @@ import { Input } from 'sajari/react-search-ui';
...
+
+By default `<Input />` has an additional mode `"instant"` and it is also the default value (search on change), if you want to disable it, simple change mode to `"standard"`:
+
+```jsx
+<form
+  onSubmit={(e) => {
+    e.preventDefault();
+    const formData = new FormData(e.target);
+    const value = formData.get('query');
+    console.log(value);
```
sampotts, line 61: Can we not just display it in the example?

docs/pages/search-ui/input.mdx
```diff
@@ -48,6 +48,23 @@ import { Input } from 'sajari/react-search-ui';
 </>
 ```
 
+## Standard
+
+By default `<Input />` has an additional mode `"instant"` and it is also the default value (search on change), if you want to disable it, simple change mode to `"standard"`:
+
+```jsx
+<form
+  onSubmit={(e) => {
```
sampotts, line 57: I'd suggest `event` rather than `e`.

sampotts (approved):

PR #242, opened 2020-11-24
V3: set viewtype in context

tuanddd:
## What this PR does
- [x] Add `viewType` state in the context
- [x] Hook `ViewType` up with context so it should just work without additional props
- [x] Make it so that `<Results />` can still control the appearance with prop


packages/hooks/src/SearchContextProvider/types.ts
```diff
@@ -53,6 +54,10 @@ export interface Context {
   instant: PipelineContextState;
   resultClicked: ResultClickedFn;
   paginate: PaginateFn;
+  viewType: {
+    value: ResultViewType;
+    setValue: (v: ResultViewType) => void;
+  };
```
sampotts, line 60: Should we make these properties `viewType` and `setViewType`? I think that's the pattern we've used elsewhere.

sampotts:
Sorry, I just created some merge conflicts too 🤦

sampotts (approved):

PR #245, opened 2020-11-25
V3: fix results grid view

tuanddd:
## What this PR does
- [x] Pass in `appearance` to override the default `undefined` coming from `props`
- [x] Increase the gap value of grid view from 4 to 8

### DIFF
--- a/packages/search-ui/src/Results/styles.ts
+++ b/packages/search-ui/src/Results/styles.ts
@@ -88,7 +88,7 @@ export default function useResultsStyles({
 
     // TODO: These sort of helpers should go in utils ideally, or find a better way to handle this
     let columns: ColumnValue = 1;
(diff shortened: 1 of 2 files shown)


sampotts (approved):

packages/search-ui/src/Results/styles.ts
```diff
@@ -88,7 +88,7 @@ export default function useResultsStyles({
 
     // TODO: These sort of helpers should go in utils ideally, or find a better way to handle this
     let columns: ColumnValue = 1;
-    let gap: GapValue = 4;
+    let gap: GapValue = 8;
```
sampotts, line 91: I think we also need to do it in the default switch case. (line 197)

sampotts (approved):

PR #251, opened 2020-11-26
V3: refactor laggy input

tuanddd:


packages/hooks/src/usePagination/index.ts
```diff
@@ -19,17 +19,22 @@ function usePagination(key: 'search' | 'instant'): UsePaginationResult {
...
+      totalResults,
+      setPage: paginate,
+      pageSize: resultsPerPage,
+      hasNext,
+      hasPrevious,
+      nextPage,
+      previousPage,
+    }),
+    [page, totalPages, totalResults, paginate, resultsPerPage, hasNext, hasPrevious, nextPage, previousPage],
+  );
```
sampotts, line 35: What is this doing exactly? I'm not sure this will save much performance-wise, will it? Creating the object should be fast?

packages/hooks/src/useQuery/index.ts
```diff
@@ -15,7 +15,9 @@ function useQuery() {
     [search, variables],
   );
 
-  return { query, setQuery };
+  const value = useMemo(() => ({ query, setQuery }), [query, setQuery]);
+
+  return value;
```
sampotts, line 20: As above. Not sure this saves much, if anything?

packages/hooks/src/useSearchContext/index.ts
```diff
@@ -13,24 +14,46 @@ function useSearchContext<T = Record<string, string | string[]>>() {
...
+      totalResults,
+      pageCount,
+      setPage,
+      search,
+      mapResponse,
+      searching,
+      viewType,
+      setViewType,
+    ],
+  );
```
sampotts, line 54: Same. Also since React doesn't do deep comparisons of objects or arrays, it'll always treat any update as an invalidation.

packages/hooks/src/useTracking/index.ts
```diff
@@ -7,10 +9,15 @@ function useTracking(): UseTrackingResult {
...
-  };
+
+  const value = useMemo(
+    () => ({
+      consumeInteractionToken: client?.interactionConsume,
+      handleResultClicked: resultClicked,
+    }),
+    [resultClicked],
+  );
+  return value;
```
sampotts, line 20: Same thing. Also could do an early return if we keep the memo.

packages/hooks/src/usePagination/index.ts
tuanddd: I was testing out if this really helps, but apparently they don't, gonna revert them and those below.

sampotts: Sorry, I was checking on my phone and didn't notice it was draft 👍

sampotts (approved):

PR #254, opened 2020-11-27
chore: move result mapping to search ui package

tuanddd:
## What this PR does
- [x] Move the mapping logic to `search-ui` package
- [x] Correct some places in doc


packages/hooks/src/useSearch/index.ts
```diff
@@ -37,10 +36,9 @@ function useCustomSearch({ pipeline, variables, fields = {} }: UseSearchCustomCo
     return pipeline.listen(EVENT_RESPONSE_UPDATED, (response: Response) => {
       const { time: latency, totalResults } = { time: response?.getTime(), totalResults: response?.getTotalResults() };
       const reqResults = response?.getResults();
```
sampotts, line 38: Can be just `const results = response?.getResults();` and then the reference below can be just `results,`. Cleaner 👍

sampotts (approved):

PR #255, opened 2020-11-28
V3: improve performance

tuanddd:
## What this PR does
- [x] Making `Combobox` a controlled by passing `inputValue`
- [x] Remove use of debouncing search functions (keeping the definition as maybe we'll need it futher down the road)
- [x] Make sure the `_id != ""` is included
- [x] Some minor optimization

### DIFF
--- a/packages/hooks/src/useFilter/index.ts
+++ b/packages/hooks/src/useFilter/index.ts
@@ -1,6 +1,6 @@
 import { isNumber } from '@sajari/react-sdk-utils';
 import { CountAggregate } from '@sajari/sdk-js';
-import { useCallback, useEffect, useMemo, useState } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 
 import { useContext } from '../SearchContextProvider';
 import { EVENT_SELECTION_UPDATED } from '../SearchContextProvider/events';
@@ -20,7 +20,6 @@ function useFilter(name: string) {
   }
 
   const [selected, setInternalSelected] = useState(filter.get());
-  const [options, setOptions] = useState<FilterItem[]>([]);
 
   useEffect(() => {
     const removeListener = filter.listen(EVENT_SELECTION_UPDATED, () => {
@@ -32,11 +31,17 @@ function useFilter(name: string) {
     };
   }, []);
 
-  useEffect(() => {
-    let output: FilterItem[] = [];
+  const setSelected = (value: string[], merge: boolean = false) => {
+    filter.set(value, merge);
+  };
 
+  const reset = () => {
+    filter.reset();
+  };
+
+  const options: FilterItem[] = useMemo(() => {
     if (!response || searching) {
-      return;
+      return [];
     }
 
     const aggregates = response.getAggregates();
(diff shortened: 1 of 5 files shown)


packages/hooks/src/useFilter/index.ts
```diff
@@ -32,11 +31,17 @@ function useFilter(name: string) {
...
 
-  useEffect(() => {
-    let output: FilterItem[] = [];
+  const setSelected = useCallback((value: string[], merge: boolean = false) => {
+    filter.set(value, merge);
+  }, []);
 
+  const reset = useCallback(() => {
+    filter.reset();
+  }, []);
```
sampotts, line 40: What's the benefit of this being in a `useCallback` if there's no dependencies?

packages/hooks/src/useFilter/index.ts
```diff
@@ -32,11 +31,17 @@ function useFilter(name: string) {
     };
   }, []);
 
-  useEffect(() => {
-    let output: FilterItem[] = [];
+  const setSelected = useCallback((value: string[], merge: boolean = false) => {
+    filter.set(value, merge);
+  }, []);
```
sampotts, line 36: As below

sampotts (approved):
Might be good for Thanh to check this too 👍

sampotts:
I think we'll still need some throttling or denouncing of the search when in instant search mode. Otherwise it'll usually result in laggy input because it's re-rendering as you type.

tuanddd:
> I think we'll still need some throttling or debouncing of the search when in instant search mode. Otherwise it'll usually result in laggy input because it's re-rendering as you type.

Ah yes but I want to explain something first:

Apparently debouncing the **whole** search function is the cause to the weird bug that I've mentioned on Friday, because it was delaying setting the query state so the result is some characters dissapeared while typing.

So in the latest change I keep everything as it is while just debouncing the actual part when the search request is sent so the state is kept in sync with the UI.

One last step is that I need to rebase the bestbuy demo branch on top of this to test out my theory but I think everything would check out eventually.

Cheers 🙂

docs/pages/search-ui/summary.mdx
```diff
@@ -49,7 +49,7 @@ function Example() {
 
 ```jsx
 function Example() {
-  const { search, searchInstant } = useSearch({ allowEmptySearch: false });
+  const { search, searchInstant } = useSearch();
```
zlatanpham, line 52: Previously, anytime the `useSearch` was declared, it will automatically trigger a search for an empty query which is not favorable in all cases. The option makes this behavior optional so we might need an alternative here if you want to remove `allowEmtpySearch`. Maybe we can consider hoisting `searchOnLoad` to the context level?

tuanddd: For the sake of this PR I'm gonna revert the change as I still haven't decided what is the best way and even if there is the fix is probably very much out of scope of this PR 🤔, I'll open another one after this. Thanks for your suggestion though 🙏

sampotts (approved):

PR #257, opened 2020-11-30
V3: improve key mapping for mapStyles util

zlatanpham:
Use the power of generic to retain the keys of `styles` object for the output of `mapStyles` function for better code autocompletion and let TS keep an eye on `undefined` check.

Before:

image

After:

image

### DIFF
--- a/packages/search-ui/src/Results/components/Result/styles.ts
+++ b/packages/search-ui/src/Results/components/Result/styles.ts
@@ -6,7 +6,7 @@ import { ResultProps } from './types';
 export default function useResultStyles(props: ResultProps) {
   const { appearance } = props;
 
-  const styles: Record<string, TwStyle[]> = {
+  const styles: Record<'container' | 'imageContainer' | 'image', TwStyle[]> = {
     container: [],
     imageContainer: [],
     image: [tw`rounded-md`],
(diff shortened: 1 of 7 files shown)


packages/search-ui/src/Results/components/Result/styles.ts
```diff
@@ -6,7 +6,7 @@ import { ResultProps } from './types';
 export default function useResultStyles(props: ResultProps) {
   const { appearance } = props;
 
-  const styles: Record<string, TwStyle[]> = {
+  const styles: Record<'container' | 'imageContainer' | 'image', TwStyle[]> = {
```
sampotts, line 9: Is there no way to make this dynamic?

zlatanpham: I can think of a way but it will require an extra function to infer the keys:

```ts
function inferTypes<T, K = TwStyle[]>(obj: T) {
  return (obj as unknown) as Record<keyof T, K>;
}

export default function useResultStyles(props: ResultProps) {
  const { appearance } = props;

  const styles = inferTypes({
    container: [],
    imageContainer: [],
    image: [tw`rounded-md`],
  });
```

PR #258, opened 2020-11-30
V3: simplify hooks logic

tuanddd:
## Why
At first it seemed convenient that we search whenever the parameters changed but on second thought, hooks should be minimal and straighforward without side effect - those effects should go in the `search-ui` package instead
## What this PR does
- [x] Remove call to `search` on hooks
- [x] Add `useEffect` inside some compositions


sampotts:
Does this mean if you call the `setPageSize` in the `usePageSize` hook, for example, it won't perform a search?

tuanddd:
> Does this mean if you call the `setPageSize` in the `usePageSize` hook, for example, it won't perform a search?

Yes you're correct

PR #259, opened 2020-12-01
feat: added TabFilter component

sampotts:
Adds support for `<Filter type="tabs">`. 

image

### DIFF
--- a/packages/search-ui/src/Filter/TabFilter.tsx
+++ b/packages/search-ui/src/Filter/TabFilter.tsx
@@ -28,8 +28,8 @@ const TabFilter = (props: Omit<TabFilterProps, 'type'>) => {
     >
       <TabList>
         <Tab>{t('filter.all')}</Tab>
-        {sliced.map(({ label, count }) => (
-          <Tab>{`${label} (${count.toLocaleString()})`}</Tab>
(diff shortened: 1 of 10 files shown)


packages/search-ui/src/Filter/TabFilter.tsx
```diff
@@ -0,0 +1,40 @@
...
+  return (
+    <Tabs
+      aria-label={title}
+      onChange={(index) => setSelected(index !== 0 ? [sliced[index - 1].label] : [])}
+      index={selectedIndex < 0 ? 0 : selectedIndex + 1}
+    >
+      <TabList>
+        <Tab>{t('filter.all')}</Tab>
+        {sliced.map(({ label, count }) => (
+          <Tab>{`${label} (${count.toLocaleString()})`}</Tab>
```
zlatanpham, line 32: Don't we need `key` prop here? It seems there is no warning on the console.

zlatanpham (approved):

packages/search-ui/src/Filter/TabFilter.tsx
sampotts: Good call. Will add it in anyway.

PR #260, opened 2020-12-01
feat: add option to control styles for ViewType

zlatanpham:
Add options to control styles for ViewType component:

- [x] custom classNames at the context level
- [x] be able to completely disable the default styling

### DIFF
--- a/packages/search-ui/src/ViewType/types.ts
+++ b/packages/search-ui/src/ViewType/types.ts
@@ -1,6 +1,6 @@
 import { BoxProps } from '@sajari/react-components';
 
-export interface ViewTypeProps extends BoxProps {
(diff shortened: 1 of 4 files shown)


sampotts (approved):

packages/search-ui/src/ViewType/types.ts
```diff
@@ -1,6 +1,6 @@
 import { BoxProps } from '@sajari/react-components';
 
-export interface ViewTypeProps extends BoxProps {
+export interface ViewTypeProps extends Omit<BoxProps, 'className'> {
```
zlatanpham, line 3: Because we have a `container` className passed down from the context so I think we can omit the option for `className` prop.

PR #264, opened 2020-12-01
V3: fix filter bucket usage

tuanddd:
## What this PR does
- [x] Update doc for bucket filter to work correctly

I think this is because there was a PR that made the `count` field `true` by default, so we forgot to check back all filters.

### DIFF
--- a/docs/pages/search-ui/filter.mdx
+++ b/docs/pages/search-ui/filter.mdx
@@ -35,6 +35,7 @@ function Example() {
       HP: "brand = 'HP'",
       Garmin: "brand = 'Garmin'",
     },
+    count: false,
   });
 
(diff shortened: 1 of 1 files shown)


tuanddd:
@sampotts you might see zeroes in the preview link but if you inspect the response payload the buckets are working except it isn't returning any `count` field back (it did when I was developing).

sampotts:
Does the `count` always have to be false if we're specifying `options` (buckets)? If that's the case then why don't we do that automatically?

sampotts:
@tuanddd I'll close this one as what I've done instead is `count: isEmpty(options)` in the `Filter` constructor instead. I think there are still some issues with this as counts for the brand bucket is empty but we can address those later.

PR #266, opened 2020-12-01
V3: reset to page 1 when apply filter

tuanddd:
## What this PR does
- [x] Move the `searchFn` function declaration up and delegate the callback for when a filter is applied to `searchFn`
- [x] Create a basic example with `search-ui` compositions

### DIFF
--- a/docs/pages/examples/basic.mdx
+++ b/docs/pages/examples/basic.mdx
@@ -1,0 +1,104 @@
+import SEO from '../../components/SEO';
+
+<SEO title="Examples - Basic" description="Basic example of using the SDK" />
+
+# Basic
+
+```jsx
+function Example() {
+  const pipeline = new Pipeline(
+    {
+      account: '1594153711901724220',
+      collection: 'bestbuy',
+      endpoint: 'https://jsonapi-us-valkyrie.sajari.net',
+    },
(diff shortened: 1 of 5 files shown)


docs/pages/examples/basic.mdx
```diff
@@ -0,0 +1,104 @@
...
+    array: true,
+  });
+
+  const ratingFilter = new FilterBuilder({
+    name: 'rating',
+    field: 'rating',
+  });
+
+  const SearchPlayground = React.memo(() => (
+    <div className="flex flex-col space-y-3">
```
sampotts, line 55: What's the reason behind `flex flex-col` here? Not a big issue but I don't think we need it?

docs/pages/examples/basic.mdx
```diff
@@ -0,0 +1,104 @@
...
+          <div className="sticky bottom-0 p-6">
+            <Pagination />
+          </div>
+        </div>
+      </div>
+    </div>
+  ));
+
+  return (
+    <SearchContextProvider
```
sampotts, line 92: For this to be a copy and paste example, I think this would be `ContextProvider` wouldn't it?

sampotts (approved):

PR #267, opened 2020-12-02
V3: add option to control styles for Swatch

zlatanpham:
Add options to control styles for Swatch component:

- [x] control classNames of `Color` component via the parent `Swatch`.
- [x] be able to completely disable the default styling.

### DIFF
--- a/packages/components/src/Swatch/styles.tsx
+++ b/packages/components/src/Swatch/styles.tsx
@@ -1,0 +1,45 @@
+import { mapStyles } from '@sajari/react-sdk-utils';
+import tw from 'twin.macro';
+
+import { useFocusRingStyles } from '../hooks';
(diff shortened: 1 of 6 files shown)


packages/components/src/Swatch/styles.tsx
```diff
@@ -0,0 +1,45 @@
...
+      styles.container.push(tw`rounded-lg`);
+      break;
+    case 'full':
+      styles.container.push(tw`rounded-full`);
+      break;
+    default:
+    case 'md':
+      styles.container.push(tw`rounded-md`);
+      break;
+  }
```
sampotts, line 42: Should we make this a `useBorderRadius` hook perhaps?

sampotts (approved):

PR #269, opened 2020-12-02
V3: fix slow render time in Swatch and Pagination

tuanddd:


packages/search-ui/src/Pagination/index.tsx
```diff
@@ -12,6 +12,16 @@ const Pagination = (props: PaginationProps) => {
...
+  const i18n = useMemo(
+    () => ({
+      label: t('pagination.label'),
+      previous: t('pagination.previous'),
+      next: t('pagination.next'),
+      page: t('pagination.page'),
+      current: t('pagination.current'),
+    }),
+    [t],
+  );
```
sampotts, line 24: This should be fast so I'm not sure we need a memo here?

tuanddd: Maybe, I'll revert the change and profile again, because it was passing new object on every render so I thought to memo it.

tuanddd: There are some noticeable performance gain (not too significant though but still), what do you think?

sampotts: I just wonder if it's going to break the functionality of the i18n stuff.

tuanddd: Understood, reverting back, we can always come back to this :)

sampotts (approved):

PR #281, opened 2020-12-05
V3: add plugin to autogenerate props table

tuanddd:
## Why
Because manually documenting props is a tedious task.

## What this PR does
- [x] Add a private package `remark-props-table` that allows us to hook into the mdx transform and inject a component's props table into the file.

## How to enable it
Inside any `.mdx` file, simply insert a `## Props Table` at the end of the file and the plugin will pick it up

## Caveats
* Apparently the docgen parser couldn't parse props where `PropsWithAs` is involved, take for example `Button` and `Image` component.
* Default props value can't also be picked up due to the face that we're using a function with default param syntax. One way we could do is to add `Component.defaultProps = {}`, that way the default props value will be populated.

### DIFF
--- a/docs/next.config.js
+++ b/docs/next.config.js
@@ -1,5 +1,24 @@
+const path = require('path');
+const fs = require('fs');
 /* eslint-disable global-require */
 const remarkPlugins = [
+  [
+    require('remark-props-table'),
+    {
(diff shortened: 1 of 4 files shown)


sampotts:
It'd be good to get this working on the `master` branch at some point.

PR #282, opened 2020-12-06
V3: rating filter improvement

tuanddd:
## Why
`ListFilter` is rerendering multiple times because of anonymous function `itemRender` -> poor performance.

## What this PR does
- [x] Wrap the render prop in `useCallback` to avoid unnecessary rerenders

### DIFF
--- a/packages/search-ui/src/Filter/RatingFilter.tsx
+++ b/packages/search-ui/src/Filter/RatingFilter.tsx
@@ -1,5 +1,5 @@
 import { Rating } from '@sajari/react-components';
-import React from 'react';
+import React, { useCallback } from 'react';
 
 import { useSearchUIContext } from '../ContextProvider';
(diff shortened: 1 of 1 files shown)


packages/search-ui/src/Filter/RatingFilter.tsx
```diff
@@ -23,9 +19,18 @@ const RatingFilter = ({ name, title }: Omit<RatingFilterProps, 'type'>) => {
           value={Number(v)}
           disableDefaultStyles={disableDefaultStyles}
         />
-      )}
-    />
+      );
+    },
+    [
+      ratingMax,
+      customClassNames.filter?.rating?.container,
```
sampotts, line 26: I guess this could be `JSON.stringify(customClassNames.filter?.rating)` ?

sampotts (approved):

PR #297, opened 2020-12-16
Add test for button group

zlatanpham:
- [x] add test for ButtonGroup
- [x] add a util wrapper for test render and ThemeProvider

### DIFF
--- a/packages/components/src/Button/test/__snapshots__/Button.test.tsx.snap
+++ b/packages/components/src/Button/test/__snapshots__/Button.test.tsx.snap
@@ -1,0 +1,71 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Button Should match snapshot 1`] = `
+<DocumentFragment>
+  .emotion-0 {
+  position: relative;
+  -webkit-align-items: center;
+  -webkit-box-align: center;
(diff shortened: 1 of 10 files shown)


packages/components/src/Button/test/__snapshots__/Button.test.tsx.snap
```diff
@@ -0,0 +1,71 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
sampotts, line 1: Will this snapshot need to be maintained or is it automated?

zlatanpham: That's when you change the structure or style of the component, the snapshot will fail which will cause the CI to fail. If you want to get the CI passed, you will need to run `yarn test -u` to update the snapshot. So basically, it means when you do some changes in the component, it makes sure you change it intentionally.

zlatanpham: We need to maintain it like https://github.com/sajari/sdk-react/pull/297/commits/7821d479926997ac2b08fee6a69d96bfb3befd3d (maybe my first answer didn't fulfill your question).

sampotts (approved):

PR #304, opened 2020-12-31
feat: new custom select and select filter type

sampotts:
Adds a `select` type for filters and a custom `Select` component with multiple select support.


packages/components/src/Select/index.tsx
```diff
@@ -1,52 +1,178 @@
...
+          <Label visuallyHidden {...getLabelProps()}>
+            {label}
+          </Label>
+        )}
+
+        <Button />
+
+        <Dropdown />
+
+        {name && !disabled && selectedItems.map((v) => <Box as="input" type="hidden" name={name} value={v} />)}
```
zlatanpham, line 173: Missing `key` prop.

zlatanpham (approved):
Really a solid one 👍. I only left a small remark.

PR #306, opened 2021-01-06
original price mapping

tuanddd:
## What this PR does
- [x] Add a new field in the `FieldDictionary` that allows for flagging item as on sale

Assuming that I understand the requirements correctly then one can do:
```js
const fields = new FieldDictionary({
  originalPrice: (data) => data.variant_compare_at_prices.map((cp, i) => (cp === '0' ? data.variant_prices[i] : cp)),
  price: (data) =>
    data.variant_prices.map((p, i) =>
      p < data.variant_compare_at_prices[i] || data.variant_compare_at_prices[i] === '0'
        ? p
        : data.variant_compare_at_prices[i],
    ),
});
```

The above is for shopify collections, the `price` will be the discount price if it is on sale, otherwise it is the same as `originalPrice`


sampotts:
Closing in favour of #369

PR #312, opened 2021-01-08
fix: list filter item has wrong focus

tuanddd:
## What this PR does
- [x] Fix the wrong focus bug in checkboxes/radios in filter list

Not sure if there's a term for this kind of bug and whether I'm handling it the right way though.

### DIFF
--- a/packages/search-ui/src/Filter/ListFilter.tsx
+++ b/packages/search-ui/src/Filter/ListFilter.tsx
@@ -1,6 +1,6 @@
 import { Box as CoreBox, Button, Checkbox, CheckboxGroup, Combobox, Radio, RadioGroup } from '@sajari/react-components';
 import { useFilter, useQuery } from '@sajari/react-hooks';
-import { getStylesObject, isBoolean, isEmpty, noop, useTheme } from '@sajari/react-sdk-utils';
+import { getStylesObject, isBoolean, isEmpty, isSSR, noop, useTheme } from '@sajari/react-sdk-utils';
 import * as React from 'react';
 import { useTranslation } from 'react-i18next';
 import tw from 'twin.macro';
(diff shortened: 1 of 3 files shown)


packages/search-ui/src/Filter/ListFilter.tsx
```diff
@@ -61,11 +63,25 @@ const ListFilter = (props: Omit<ListFilterProps, 'type'>) => {
     setShown(limit);
   }, [query, q]);
 
+  // On user interaction, get the last element being interacted with BEFORE the list is rerendered and focus that element
+  React.useEffect(() => {
+    if (!isSSR()) {
```
sampotts, line 68: Worth checking that `pinSelected` is true, otherwise this logic is not required.

packages/search-ui/src/Filter/ListFilter.tsx
```diff
@@ -61,11 +63,25 @@ const ListFilter = (props: Omit<ListFilterProps, 'type'>) => {
...
+      const input = document
+        .querySelector(`#list-${name}`)
```
sampotts, line 70: Should we set this as a constant since it's referenced in a few spots now

packages/search-ui/src/Filter/ListFilter.tsx
```diff
@@ -61,11 +63,25 @@ const ListFilter = (props: Omit<ListFilterProps, 'type'>) => {
...
+      input?.focus();
+    }
+  }, [JSON.stringify(selected)]);
+
   const Control = multi ? Checkbox : Radio;
   const filtered = searchable ? options.filter((o) => o.label.toLowerCase().includes(query.toLowerCase())) : options;
   const slice = filtered.length > limit;
 
   const sortedItems = React.useMemo(() => {
+    if (!isSSR()) {
```
sampotts, line 82: Again, maybe we want to check `pinSelected` here?

sampotts (approved):

PR #313, opened 2021-01-08
fix: make RangeFilter more failsafe default for non aggregate

zlatanpham:


packages/hooks/src/ContextProvider/controllers/filters/RangeFilterBuilder.ts
```diff
@@ -28,14 +28,19 @@ export default class RangeFilterBuilder {
...
-    max = 0,
     aggregate = true,
+    initial,
+    min = 0,
+    max = aggregate ? 0 : 100,
     formatter = (value: Range) => value.map(Math.round) as Range,
   }: RangeFilterOptions) {
-    this.range = initial;
-    this.initial = initial;
+    if (initial === undefined) {
```
sampotts, line 37: Should we use `typeof initial === 'undefined'` ?

sampotts (approved):

PR #318, opened 2021-01-10
fix: fix pagination aria labels

sampotts:


zlatanpham (dismissed):
Just need to update the failing test, otherwise, LGTM.

PR #323, opened 2021-01-11
feat: added support for ARRAY_MATCH filter grouping

sampotts:


tuanddd (approved):
LGTM, just push a new doc section for `ARRAY_MATCH` and you're good to go.

packages/hooks/src/ContextProvider/controllers/filters/utils.ts
```diff
@@ -0,0 +1,69 @@
...
+    ) as Record<string, Array<string>>;
+
+  // Build filter expression
+  if (type === 'filter') {
+    // Build the list of non grouped filters
+    const basicFilters = list.filter(({ group, count }) => !group && !count).map(({ expression }) => `(${expression})`);
+
+    return Object.entries(groups)
+      .reduce((out, [group, expressions]) => {
+        if (group !== 'undefined' && !isEmpty(expressions)) {
```
zlatanpham, line 47: Should it be `typeof group !== 'undefined'`?

zlatanpham (approved):
Aside from one minor comment, it looks good 👍

packages/hooks/src/ContextProvider/controllers/filters/utils.ts
sampotts: Good catch but due to the way the groupBy function works, if `group` is `undefined` in the `FilterBuilder` then it gets turned into a string for the key of the object so it ends up as `'undefined'` instead. Hopefully, that makes sense 👍

PR #336, opened 2021-01-12
fix: revert server default filter

tuanddd:


packages/server/src/search/index.tsx
```diff
@@ -1,28 +1,26 @@
...
-  });
+  const { pipeline, filters, variables = new Variables() } = props;
+
+  if (filters) {
+    const filter = combineFilters(filters);
+
+    variables.set({
+      filter: () => {
+        const expression = filter.filter();
+        return [isEmpty(expression) ? '_id != ""' : expression].filter(Boolean).join(' AND ');
```
sampotts, line 17: I think this can be `return isEmpty(expression) ? '_id != ""' : expression;`?

sampotts (approved):

PR #358, opened 2021-01-18
fix: useFilter sorting

tuanddd:


packages/hooks/src/useFilter/index.ts
```diff
@@ -88,12 +88,14 @@ function useFilter(name: string, params: { sort?: SortType; sortAscending?: bool
       return (count[value] as number) ?? 0;
     };
 
-    return sortItems(Object.entries(filter.getOptions())).map(([label, value]) => {
-      const id = `${name}_${label}`;
-      const count = getBucketCount(id);
-
-      return { label, value, count } as FilterItem;
-    });
+    return sortItems(Object.entries(filter.getOptions()), params.sort ?? 'alpha', params.sortAscending ?? true).map(
```
sampotts, line 91: Should we make `sortAscending` default to true if sort is not 'count' like the Filter component?

sampotts (approved):

sampotts (approved):

PR #366, opened 2021-01-20
feat: create Modal component

zlatanpham:
Add a simple, accessible modal component that can be used for the overlay option in the website integration, effectively rendering the whole search UI within the modal. This work is heavily influenced by [react-responsive-modal](https://github.com/pradel/react-responsive-modal) which is small in bundle size.

- [x] Trap focus within the modal, adding `inert` to all other elements on the page. 
- [x] Close when pressing ESC
- [x] Be responsive
- [x] Update docs
- [x] Add tests


packages/components/src/Modal/lib/focusTrapJs.ts
```diff
@@ -0,0 +1,105 @@
...
+
+function isContentEditable(node: any) {
+  return node.getAttribute('contentEditable');
+}
+
+function getTabindex(node: any) {
+  const tabindexAttr = parseInt(node.getAttribute('tabindex') || '', 10);
+
+  // eslint-disable-next-line no-restricted-globals
+  if (!isNaN(tabindexAttr)) return tabindexAttr;
```
sampotts, line 24: We can use `isNumber` here can't we?

packages/components/src/Modal/lib/focusTrapJs.ts
```diff
@@ -0,0 +1,105 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+// https://github.com/alexandrzavalii/focus-trap-js/blob/master/src/index.js v1.1.0
```
sampotts, line 2: Did you consider using a `FocusLock` component as we have in Sajari UI?
https://github.com/sajari/sajari-ui/blob/master/packages/sajari-ui/src/components/FocusLock/index.tsx

sampotts (commented):
I've left a couple of comments on the focus lock stuff

packages/components/src/Modal/lib/focusTrapJs.ts
zlatanpham: I've just pushed the refactor.

sampotts (approved):

PR #380, opened 2021-02-01
feat: add attach mode to combobox

tuanddd:


packages/components/src/Combobox/index.tsx
```diff
@@ -238,6 +240,20 @@ const Combobox = React.forwardRef(function ComboboxInner<T>(props: ComboboxProps
...
+      throw new Error('When using `attach` mode you must pass the `inputElement` prop.');
+    }
+    const container = inputElement.parentElement;
+    ReactDOM.render(
+      <ComboboxContextProvider value={context}>
+        <Dropdown />
+      </ComboboxContextProvider>,
+      container,
+    );
+    return null;
```
sampotts, line 254: So won't this just render over the top of the old input? I think ideally we want to retain the styling of the current input, if possible. This would wipe it out and replace it with ours I think.

packages/components/src/Combobox/types.ts
```diff
@@ -88,6 +88,8 @@ interface Props<T> {
     selected: boolean;
     getItemProps: UseComboboxReturnValue<T>['getItemProps'];
   }) => React.ReactNode;
+  /** The input element, used in conjuntion with mode `attach` */
+  inputElement?: HTMLInputElement;
```
sampotts, line 92: I'm just wondering if we can determine that it's in "attach mode" from this prop being set? That way they can also specify `'suggestions'`, `'results'` or `'standard'` modes?

sampotts:
It'd be good to have some docs for this with an example?

sampotts (dismissed):

sampotts:
Hey @tuanddd, sorry, could you fix the merge conflict on this and I'll get it merged.

sampotts (approved):

PR #394, opened 2021-02-08
fix: improve usability of RangeInput component

sampotts:
Previously, the RangeInput manual inputs were finicky to work with. This was because we were always clamping the value to the closest step on input, rather than waiting for the user to hit the return key which seems to be the common pattern for this type of input. 

Before:

https://user-images.githubusercontent.com/719092/107172222-e6ddff80-6a18-11eb-8661-8a37cdf19362.mp4

After:

https://user-images.githubusercontent.com/719092/107172968-cdd64e00-6a1a-11eb-822c-fbef88845b78.mp4

Internal issue: https://sajari.atlassian.net/jira/software/projects/SF/boards/29?selectedIssue=SF-244


zlatanpham (approved):
LGTM 👍

PR #396, opened 2021-02-08
utils: add MergeOptions for merge util

sampotts:
This change adds the ability to specify how arrays are merged. Currently, the options are:

 * `'concat'` - Concatenate arrays.
 * `'union'` - Union arrays, skipping items that already exist.
 * `'replace'` - Replace all array items.

### DIFF
--- a/packages/utils/src/object.ts
+++ b/packages/utils/src/object.ts
@@ -18,16 +18,15 @@ interface MergeObject {
   [k: string]: any;
 }
 
-/**
- * How arrays should be merged
- * 'concat' Concatenate arrays.
- * 'union' Union arrays, skipping items that already exist.
- * 'replace' Replace all array items.
- */
 type MergeArrayHandling = 'concat' | 'union' | 'replace';
 
 export class MergeOptions {
-  /** Whether to concatinate arrays */
+  /**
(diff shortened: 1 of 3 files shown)


packages/utils/src/object.ts
```diff
@@ -18,12 +18,40 @@ interface MergeObject {
...
+    this.arrayHandling = options?.arrayHandling ?? 'concat';
+  }
+}
+
 // Deep extend destination object with N more objects
-export function merge(target: MergeObject, ...sources: MergeObject[]): MergeObject {
+export function merge(
+  options: MergeOptions | MergeObject,
+  target: MergeObject,
+  ...sources: MergeObject[]
```
zlatanpham, line 41: It's a bit unusual to me to have the options as the first argument but not sure if we can do anything better since we rely on the function rest parameters.

zlatanpham (approved):
I've left one comment but I don't know if it's worth an update so I'll let you decide.

packages/utils/src/object.ts
sampotts: Yeah, it's a bit of an odd API I'll admit but I handle it if it's omitted.

PR #414, opened 2021-02-18
fix: typing mid query snaps to the end of the query

zlatanpham:
Fix the issue raised internally by the team [here](https://sajari.atlassian.net/browse/SF-250). A similar issue was raised and fixed in https://github.com/downshift-js/downshift/issues/1108.


sampotts (approved):

sampotts:
Nice one! I spent a while trying to get to the bottom of this! 👍

PR #429, opened 2021-03-03
feat: enable image resizing

tuanddd:


packages/components/src/Image/useImageProvider.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  switch (provider) {
+    case 'shopify':
+      return images.map((i) => {
+        const url = new URL(i);
+        const { pathname } = url;
+        const [extension, ...rest] = pathname.split('.').reverse();
+        const pathnameWithoutExtension = rest.join();
+        if (width && height) {
+          url.pathname = `${pathnameWithoutExtension}_${String(width)}x${String(height)}`;
```
sampotts, line 27: If I'm not mistaken, the `width` and `height` values will automatically get converted to string in a template literal.  

Same for below.

sampotts (commented):
This is looking awesome. Just one small comment from me.

sampotts (approved):

PR #430, opened 2021-03-04
fix: add relative position for input container

tuanddd:
@sampotts Sorry I think the relative positioning should have been here instead of in the widgets.


sampotts (dismissed):

sampotts:
@tuanddd Do we need to bump the `search-ui` package too?

PR #441, opened 2021-03-18
feat: support maxSuggestions for Input

zlatanpham:


tuanddd (approved):

tuanddd:
Should we update the markdown as well?

packages/search-ui/src/Input/index.tsx
```diff
@@ -24,8 +24,7 @@ const Input = React.forwardRef((props: InputProps<any>, ref: React.Ref<HTMLInput
   if (mode === 'suggestions') {
     items = suggestions;
```
tuanddd, line 25: Do we also need to apply to this mode?

zlatanpham: Nice catch! I've just pushed the update.

tuanddd: We should also update the markdown doc props table. Cheers

tuanddd (dismissed):

tuanddd (approved):

PR #444, opened 2021-03-23
fix: voice input not trigger search

zlatanpham:
Fix the issue when using speech search does not trigger a search for the voice input. Also, make an update to open the dropdown menu if the input mode is `results`.

**Standard mode**

https://user-images.githubusercontent.com/12707960/112083180-29b6f980-8bb9-11eb-948a-367808266a11.mov

**Results mode**

https://user-images.githubusercontent.com/12707960/112082668-56b6dc80-8bb8-11eb-98ca-4227b5e39abd.mov

### DIFF
--- a/.changeset/five-waves-press.md
+++ b/.changeset/five-waves-press.md
@@ -1,0 +1,7 @@
+---
+'@sajari/react-components': patch
+'@sajari/react-search-ui': patch
+'sajari-sdk-docs': patch
+---
+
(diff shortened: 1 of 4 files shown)


tuanddd (approved):

tuanddd:
Might be worth to put an example on the Input doc page?

zlatanpham:
> Might be worth to put an example on the Input doc page?

@tuanddd I think it's unnecessary as we explicitly said the Input inherits the props from the Combobox and we already had the example for speech recognition in the page.

PR #454, opened 2021-04-12
fix: modal being overlapped by elements having higher z-index

zlatanpham:
Increase the `z-index` value of `Modal` to 10000 to somewhat avoid it being overlapped by other elements.

### DIFF
--- a/packages/utils/src/styles/tailwind.config.js
+++ b/packages/utils/src/styles/tailwind.config.js
@@ -34,7 +34,7 @@ export default {
         inherit: 'inherit',
       },
       zIndex: {
-        10000: 1000,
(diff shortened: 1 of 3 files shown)


packages/utils/src/styles/tailwind.config.js
```diff
@@ -33,6 +33,9 @@ export default {
       fontFamily: {
         inherit: 'inherit',
       },
+      zIndex: {
+        10000: 1000,
```
tuanddd, line 37: Missing a `0`.

tuanddd (dismissed):

tuanddd (approved):

PR #460, opened 2021-04-14
feat: support reset all filters

zlatanpham:
Resolve https://github.com/sajari/sdk-react/issues/451


packages/hooks/src/ContextProvider/controllers/filters/RangeFilterBuilder.ts
```diff
@@ -162,7 +162,7 @@ export default class RangeFilterBuilder {
     if (this.frozen) {
       return;
     }
-    this.range = isArray(this.initial) ? [...this.initial] : this.initial;
+    this.range = isArray(this.initial) ? [...this.initial] : this.initial || [this.min, this.max];
```
tuanddd, line 165: Why would we need this fallback?

zlatanpham: If the values of `min` and `max` come from aggregation, we should reset the current range value to the `[min, max]` otherwise, the UI will be hidden since the `initial` is always `null`. For better, I think I will need to add a check for that.

tuanddd (approved):

PR #461, opened 2021-04-14
feat: add option to transform text in filter options

zlatanpham:
Support options for transforming text for `list`, `select` and `tab` filter


zlatanpham:
@tuanddd just hold up reviewing if you are going to do it. I think I need to add the functionality for other types of `Filter` as well.

packages/search-ui/src/Filter/TabFilter.tsx
```diff
@@ -44,15 +45,15 @@ const TabFilter = (props: Omit<TabFilterProps, 'type'>) => {
           className={customClassNames.filter?.tabs?.tab}
           selectedClassName={customClassNames.filter?.tabs?.selectedTab}
         >
-          {t('all')}
+          {formatLabel('all', { textTransform, t })}
```
tuanddd, line 48: Screen Shot 2021-04-15 at 12 59 12
Screen Shot 2021-04-15 at 13 03 10
Difference between [new change](https://deploy-preview-461--sajari-react-sdk-v3.netlify.app/search-ui/filter#tabs) and [production](https://react.docs.sajari.com/search-ui/filter#tabs)

I think this should be
```suggestion
          {formatLabel(t('all'), { textTransform })}
```

tuanddd (changes requested):
I thought we agreed that the default text transform is capitalize?

packages/search-ui/src/Filter/TabFilter.tsx
zlatanpham: Fixed but we still need to pass `t` in.

zlatanpham:
> I thought we agreed that the default text transform is capitalize?

we can do it in the `search-widget`.

tuanddd (approved):

PR #471, opened 2021-04-22
fix: remove focus border around input

tuanddd:


zlatanpham (commented):
I don't think we should permanently remove the border as it's more to turn the style off in some cases like for the overlay mode. So I think we might need to make it customizable instead.

PR #486, opened 2021-05-10
feat: add list display for sorting component

zlatanpham:
Add `list` display option for sorting component. This variant will be used for the mobile version of the overlay mode.

image

### DIFF
--- a/docs/pages/search-ui/sorting.mdx
+++ b/docs/pages/search-ui/sorting.mdx
@@ -74,7 +74,7 @@ function Example() {
 
 ## Display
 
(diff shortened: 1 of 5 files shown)


docs/pages/search-ui/sorting.mdx
```diff
@@ -72,11 +72,67 @@ function Example() {
 }
 ```
 
+## Display
+
+The sorting component comes in 2 variants: `inline` and `list`. Pass the variant prop and set it to one of these values.
```
tuanddd, line 77: Should it be `select` and `list` instead?

tuanddd (approved):

PR #506, opened 2021-06-01
BREAKING CHANGE: update ResultClicked

tuanddd:
Needs testing


.changeset/giant-grapes-clean.md
```diff
@@ -0,0 +1,16 @@
+---
+'@sajari/react-hooks': major
+'@sajari/react-search-ui': major
+---
+
+Update ResultClickedFn to receive more info
+
+Before:
+pipeline.listen('result-clicked', data => {
```
jonathaningram, line 9: Surround in code blocks and might be more clear about what data is in the `After:` one if you just deference it.

i.e. Name `data` `url` in the before and deference it in the after.

packages/hooks/.babelrc
```diff
@@ -0,0 +1,3 @@
+{
```
jonathaningram, line 1: nit: this file does not seem to be formatted?

packages/hooks/src/ContextProvider/controllers/Pipeline.test.ts
```diff
@@ -0,0 +1,45 @@
...
+    const responseUpdated = jest.fn((response) => response);
+
+    pipeline.listen(EVENT_RESPONSE_UPDATED, responseUpdated);
+    pipeline.search({ q: 'some search string' });
+    await waitFor(() => expect(responseUpdated).toHaveBeenCalled());
+
+    expect(responseUpdated.mock.results[0].value.isEmpty()).toBeFalsy();
+  });
+
+  it('should emit result clicked listern with the right arg', () => {
```
jonathaningram, line 26: listener?

packages/hooks/src/ContextProvider/controllers/Pipeline.ts
```diff
@@ -129,9 +130,9 @@ export class Pipeline {
    * Emits a result clicked event to the results clicked event listeners.
    * @param value Value to send to the listeners.
    */
-  public emitResultClicked(value: string): void {
+  public emitResultClicked(args: { token: string; values: ResultValues }): void {
```
jonathaningram, line 133: Doccomment is out of date.

packages/hooks/src/ContextProvider/types.ts
```diff
@@ -60,9 +60,21 @@ export interface Context {
...
+  image?: string | Array<string>;
+  rating?: number;
+  price?: string | Array<string>;
+  originalPrice?: string | Array<string>;
+}
+
 type Field = ((data: Record<string, any>) => any) | string | string[] | false;
 export class FieldDictionary {
-  id?: Field;
+  sajariUUID?: Field;
```
jonathaningram, line 77: How about just _id since that's what it is?

Alternatively if we want to use a camelCase name how about `internalId`?

packages/hooks/src/ContextProvider/types.ts
```diff
@@ -60,9 +60,21 @@ export interface Context {
...
+  image?: string | Array<string>;
+  rating?: number;
+  price?: string | Array<string>;
+  originalPrice?: string | Array<string>;
+}
+
 type Field = ((data: Record<string, any>) => any) | string | string[] | false;
 export class FieldDictionary {
-  id?: Field;
+  internalId?: Field;
```
jonathaningram, line 77: OK so previously a caller could override `id` like this:

```
new FieldDictionary({
  title: 'name',
  id: 'my_id'
})
```

Which says that we should look up the field named `my_id` in the record values everywhere we need to use an ID.

I'm not sure we need to change this here.

For Shopify, I think we just need to make sure the onclick handler receives all of the record values, `_id`, `id`, `url`, etc. Whatever record values exist, they should be passed in the handler as well as token.

packages/hooks/src/test/default-config.ts
```diff
@@ -0,0 +1,19 @@
+import { FieldDictionary, Pipeline } from '@sajari/react-hooks';
+
+export const pipeline = new Pipeline(
+  {
+    account: '1594153711901724220',
```
jonathaningram, line 5: Is this a real CID? If so it can't be baked into code. If not, can we make it something generic so it doesn't look real?

jonathaningram (approved):

PR #509, opened 2021-06-03
fix: update input component typings and documentation

tuanddd:


docs/pages/search-ui/input.mdx
```diff
@@ -250,8 +250,22 @@ function Example() {
...
+| `placeholder`      |      |         |             |
+| `onChange`         |      |         |             |
+| `onSelect`         |      |         |             |
+| `inputElement`     |      |         |             |
+| `enableVoice`      |      |         |             |
+| `className`        |      |         |             |
+| `showPoweredBy`    |      |         |             |
+| `showDropdownTips` |      |         |             |
+| `variant`          |      |         |             |
+| `size`             |      |         |             |
```
zlatanpham, line 266: Can you move the table + its above text down? Also, it'd be good if we can have a full description of the props so readers don't have to visit the docs of Combobox to check them.

Screen Shot 2021-06-03 at 15 07 23

zlatanpham (approved):
Thanks for the update.

PR #514, opened 2021-06-09
feat: variant option

tuanddd:
Add ability to show a product's variant images

It’s worth noting that since this new feature is added it will collide with the current image hover feature. To resolve that, only products that have variant images to show will no longer change on hover.

### DIFF
--- a/packages/components/src/Image/styles.ts
+++ b/packages/components/src/Image/styles.ts
@@ -45,7 +45,6 @@ function useObjectPosition(objectPosition: ImageProps['objectPosition']) {
     case 'right-bottom':
       return tw`object-right-bottom`;
     case 'center':
-      return tw`object-center`;
     default:
       return tw`object-center`;
   }
--- a/packages/components/src/Image/styles.ts
+++ b/packages/components/src/Image/styles.ts
@@ -26,22 +26,49 @@ function useObjectFit(objectFit: ImageProps['objectFit'], isSecondImage = false)
   }
 }
 
+function useObjectPosition(objectPosition: ImageProps['objectPosition']) {
+  switch (objectPosition) {
+    case 'top':
+      return tw`object-top`;
+    case 'bottom':
+      return tw`object-bottom`;
+    case 'left':
+      return tw`object-left`;
+    case 'right':
+      return tw`object-right`;
+    case 'left-top':
+      return tw`object-left-top`;
+    case 'left-bottom':
+      return tw`object-left-bottom`;
+    case 'right-top':
+      return tw`object-right-top`;
+    case 'right-bottom':
+      return tw`object-right-bottom`;
+    case 'center':
+      return tw`object-center`;
+    default:
+      return tw`object-center`;
+  }
+}
+
 export function useImageStyles(props: ImageProps & { hover: boolean }) {
   const hasLoaded = useHasImageLoaded(props);
-  const { hover, objectFit } = props;
+  const { hover, objectFit, objectPosition } = props;
   const styles = {
-    container: [tw`relative rounded-md`, !hasLoaded ? tw`bg-gray-100` : ''],
+    container: [tw`relative rounded-sm`, !hasLoaded ? tw`bg-gray-100` : ''],
     image: [
       tw`transition-opacity duration-200 ease-in`,
       hasLoaded && !hover ? tw`opacity-100` : tw`opacity-0`,
       { borderRadius: 'inherit' },
       useObjectFit(objectFit),
+      useObjectPosition(objectPosition),
     ],
     secondImage: [
       tw`absolute top-0 left-0 object-cover w-full h-full transition-opacity duration-200 ease-in`,
       hasLoaded && hover ? tw`opacity-100` : tw`opacity-0`,
       { borderRadius: 'inherit' },
       useObjectFit(objectFit, true),
+      useObjectPosition(objectPosition),
     ],
   };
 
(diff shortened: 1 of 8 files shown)


packages/hooks/src/ContextProvider/types.ts
```diff
@@ -70,6 +70,9 @@ export interface ResultValues {
   rating?: number;
   price?: string | Array<string>;
   originalPrice?: string | Array<string>;
+  imageIds?: string[];
+  variantIds?: string[];
+  variantImageIds?: string[];
```
zlatanpham, line 75: I'm wondering why it's not just one field to hold a list of images? For examples:
```suggestion
 images?: string[];
```

tuanddd: All 3 have different values and meaning, in order to get the right variant images please [see this](https://sajari.atlassian.net/wiki/spaces/SAJARI/pages/1209991336/Search+Result+Settings?focusedCommentId=1283096679#comment-1283096679)

zlatanpham: If so, I think we are opinionating Shopify settings here. Ideally, the component should know nothing regarding Shopify and render the images based on a single prop. We can compute the value of the prop via the [mapResultsField](https://github.com/sajari/sdk-react/blob/master/packages/search-ui/src/utils/mapResultFields.ts#L40), don't we?

tuanddd: You are correct, I'll make the changes.

packages/search-ui/src/Results/components/Result/index.tsx
```diff
@@ -165,22 +163,39 @@ const Result = React.memo(
...
+                {image.map((url, i) => {
+                  return (
+                    <Box
+                      css={styles.previewImageContainer}
+                      onMouseEnter={() => {
+                        setImageSrc(url);
+                        setActiveImageIndex(i);
+                      }}
+                    >
+                      <Image src={url} aspectRatio={1} objectFit="cover" objectPosition="top" />
```
zlatanpham, line 192: Often, it's helpful to support `alt` for the thumbnail pattern. Should we consider adding it?

packages/search-ui/src/Results/components/Result/index.tsx
```diff
@@ -212,9 +227,28 @@ const Result = React.memo(
...
-              {decodeHTML(description)}
-            </Text>
+            <Box css={styles.wrapper}>
+              {isArray(image) && image.length > 1 ? (
+                <Box css={styles.previewImagesContainer}>
+                  {image.map((url, i) => {
+                    return (
+                      <Box
+                        css={styles.previewImageContainer}
+                        onMouseEnter={() => {
```
zlatanpham, line 237: I don't see from the code that it supports keyboard accessibility. Can we have the support?

zlatanpham (changes requested):
@tuanddd I've left some suggestions. Furthermore, I don't see a change of the docs in the PR, can you make the update?

packages/search-ui/src/Results/components/Result/index.tsx
tuanddd: Instead of adding `alt` to the image, I added `role=img` and `aria-label` to the container.

zlatanpham (commented):
@tuanddd can we have an example in the docs for the change? Additionally, this section looks broken which might be caused by the change.

image

packages/search-ui/src/Results/components/Result/index.tsx
```diff
@@ -160,27 +158,65 @@ const Result = React.memo(
...
+    const renderPreviewImages = () => {
+      const getSetActive = useCallback(
+        (url: string, i: number) => () => {
+          setImageSrc(url);
+          setActiveImageIndex(i);
+        },
+        [],
+      );
+      return (
+        <Box css={styles.previewImagesContainer} aria-label={`${decodeHTML(title)}'s images`}>
```
zlatanpham, line 170: Can we do translation here?

packages/search-ui/src/Results/components/Result/index.tsx
```diff
@@ -160,27 +158,65 @@ const Result = React.memo(
...
+          {image &&
+            isArray(image) &&
+            image.map((url, i) => {
+              const setActive = getSetActive(url, i);
+              return (
+                <Box
+                  role="img"
+                  key={`product-image-${url}`}
+                  css={styles.previewImageContainer}
+                  aria-label={`${decodeHTML(title)} image number ${i + 1}`}
```
zlatanpham, line 180: Same question with translation.

packages/components/src/Image/styles.ts
```diff
@@ -26,22 +26,49 @@ function useObjectFit(objectFit: ImageProps['objectFit'], isSecondImage = false)
...
+    case 'left-bottom':
+      return tw`object-left-bottom`;
+    case 'right-top':
+      return tw`object-right-top`;
+    case 'right-bottom':
+      return tw`object-right-bottom`;
+    case 'center':
+      return tw`object-center`;
+    default:
+      return tw`object-center`;
```
zlatanpham, line 50:
```suggestion
    default:
      return tw`object-center`;
```

packages/search-ui/src/Results/components/Result/index.tsx
```diff
@@ -160,27 +160,68 @@ const Result = React.memo(
...
+        [],
+      );
+      return (
+        <Box
+          css={styles.previewImagesContainer}
+          aria-label={t('previewImagesContainer', { product: decodeHTML(title) })}
+        >
+          {image &&
+            isArray(image) &&
+            image.map((url, i) => {
```
zlatanpham, line 178: Shoud `isArray` be enough for the case?

```suggestion
          {isArray(image) &&
            image.map((url, i) => {
```

zlatanpham (approved):
I've left 2 more suggestions, see if you think they need to be updated. Otherwise, it looks good to merge.

zlatanpham (approved):

PR #523, opened 2021-06-21
fix: missing type exports in Variables

zlatanpham:
Fix missing type exports in Variables.


packages/hooks/src/ContextProvider/controllers/index.ts
```diff
@@ -1,4 +1,4 @@
-/* eslint-disable import/named */
+/* eslint-disable import/no-cycle */
```
zlatanpham, line 1: @tuanddd I had a hard time trying to fix the warning. Not sure if you can see how we can remove the line.

tuanddd: This needs a thorough commitment, we'll do it in another pull request.

tuanddd (approved):

PR #525, opened 2021-06-21
feat: add product statuses to result component

tuanddd:


packages/search-ui/src/Results/components/Result/styles.ts
```diff
@@ -21,14 +23,30 @@ export default function useResultStyles(props: UseResultStylesParams) {
     description: [tw`mt-1 text-sm text-gray-500`],
     rating: [tw`mt-1 space-y-1 text-center`],
     priceContainer: [],
-    price: [isOnSale ? tw`font-medium text-red-500` : ''],
```
zlatanpham, line 24: Not sure if we still need `font-medium` for `onSale`.

packages/search-ui/src/ContextProvider/types.ts
```diff
@@ -28,6 +28,9 @@ export interface SearchUIContextProviderValues {
       searchingMessage?: string;
       emptyMessage?: string;
       errorMessage?: string;
+      onSaleStatus?: string;
+      outOfStockStatus?: string;
+      newArrivalStatus?: string;
```
zlatanpham, line 33: Looks like we need to update the docs for the change.

packages/search-ui/src/Results/components/Result/index.tsx
```diff
@@ -87,7 +95,31 @@ const Result = React.memo(
...
-    const styles = getStylesObject(useResultStyles({ ...props, appearance, isOnSale }), disableDefaultStyles);
+    const isOutOfStock = React.useMemo(() => {
+      if (isEmpty(quantity)) {
+        return false;
+      }
+      const parseQuantities = (input: string | Array<string>) => (isArray(input) ? input : [input]).map(Number);
+      const quantities = parseQuantities(quantity);
+
+      return quantities[activeImageIndex] === 0;
+    }, [JSON.stringify(quantity), activeImageIndex]);
```
zlatanpham, line 106: Why do we need to `JSON.stringify`? Given we've already wrapped the `Result` component inside `React.memo` with a similar check.

packages/search-ui/src/Results/components/Result/styles.ts
tuanddd: If so I think `font-medium` should be applied to all statuses and not just `onSale` for consistency sake.

packages/search-ui/src/ContextProvider/types.ts
tuanddd: Updated

packages/search-ui/src/Results/components/Result/index.tsx
tuanddd: If might as well remove the `JSON.stringify` call at `isOnSale` memo as well?

tuanddd:
@zlatanpham I've also added the translation for all languages (only english is correct, others need to be updated)

packages/search-ui/src/Results/components/Result/index.tsx
zlatanpham: I would say so

zlatanpham (approved):
LGTM 👍

PR #526, opened 2021-06-22
fix: return empty if price not found

tuanddd:


packages/search-ui/src/Results/components/Result/index.tsx
```diff
@@ -129,11 +131,16 @@ const Result = React.memo(
...
+
+      const priceDisplay = isArray(price) ? price[activeImageIndex] : price;
+
+      if (isEmpty(priceDisplay)) return null;
+
       return (
         <Box css={styles.priceContainer}>
           <Text css={styles.price} className={priceClassName} disableDefaultStyles={disableDefaultStyles}>
-            {formatPrice(priceDisplay, { currency, language })}
+            {formatPrice(priceDisplay as string, { currency, language })}
```
zlatanpham, line 143: Why do we need the casting? can `priceDisplay` be another type instead of `string`?

packages/utils/src/assertion.ts
```diff
@@ -61,7 +61,7 @@ export const isEmptyArray = (value: any) => isArray(value) && value.length === 0
 export const isEmptyObject = (value: any) => isObject(value) && Object.keys(value).length === 0;
 
 // Empty assertions
-export const isEmpty = (value: any) => {
+export function isEmpty(value: any): value is null | undefined {
```
tuanddd, line 64: @zlatanpham Not sure if this covers all of other `isEmpty` calls but when I ran `yarn build` everything build OK and the `priceDisplay` type is `string`.

zlatanpham (approved):

PR #530, opened 2021-06-29
fix: capitalize util works incorrectly

zlatanpham:
Fix `capitalize` utility works incorrectly, where a string in uppercase letters is not converted to a capitalized output. For example, `hELLO` is converted to `HELLO` instead of `Hello`.

### DIFF
--- a/packages/search-ui/src/Filter/utils.ts
+++ b/packages/search-ui/src/Filter/utils.ts
@@ -35,7 +35,7 @@ interface FormatValueParams {
  * @returns {string} the capitalized value
  */
 export function capitalize(value: string): string {
-  return value.replace(/(^\w{1})|(\s+\w{1})/g, (letter) => letter.toUpperCase());
+  return value.toLocaleLowerCase().replace(/(^\w{1})|(\s+\w{1})/g, (letter) => letter.toUpperCase());
 }
 
 /**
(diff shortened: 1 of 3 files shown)


packages/search-ui/src/Filter/utils.ts
```diff
@@ -35,7 +35,7 @@ interface FormatValueParams {
  * @returns {string} the capitalized value
  */
 export function capitalize(value: string): string {
-  return value.replace(/(^\w{1})|(\s+\w{1})/g, (letter) => letter.toUpperCase());
+  return value.toLocaleLowerCase().replace(/(^\w{1})|(\s+\w{1})/g, (letter) => letter.toUpperCase());
```
tuanddd, line 38: Why use `toLocaleLowerCase` here? If so the latter should also be `toLocaleUpperCase` I suppose?

zlatanpham: It first converts the string to lowercase letters and then capitalizes the first letter per word.

zlatanpham: @tuanddd I misunderstood your question. Pushed the update to use `toLocaleUpperCase` instead.

tuanddd (approved):

PR #532, opened 2021-06-29
feat: add option for toggling variant image

tuanddd:


docs/pages/search-ui/results.mdx
```diff
@@ -86,12 +86,51 @@ function Example() {
...
+          },
+        }),
+        variables: new Variables({
+          resultsPerPage: 6,
+          q: 'macbook',
+        }),
+      }}
+      searchOnLoad
+    >
+      <Results />
```
zlatanpham, line 121: Can you check if the example is set correctly? Looks like it should be `<Results  showVariantImage />`

.changeset/hot-vans-wonder.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sajari/react-search-ui': minor
+---
+
+Add option for toggling variant image
```
zlatanpham, line 5:
```suggestion
Add option for toggling variant images.
```

packages/search-ui/src/Results/components/Result/types.ts
```diff
@@ -14,6 +14,8 @@ interface Props extends Pick<ResultsProps, 'appearance' | 'imageAspectRatio' | '
   onClick?: UseClickTrackingParams['onClick'];
   /** Display image or not */
   showImage?: boolean;
+  /** Display variant images or not */
+  showVariantImage?: boolean;
```
zlatanpham, line 18: Since it might be relevant to the price change, probably name it `showVariant` makes more sense?

```suggestion
  showVariant?: boolean;
```

zlatanpham (commented):
Apart from a couple of suggestions, it looks fine to me.

packages/search-ui/src/Results/components/Result/types.ts
tuanddd: I don't think so, `showVariantImage` is solely being used to show images, price change is just a result of hovering on each image, plus `showVariant` might not sound specific enough when first read (variant could be anything).

zlatanpham: But we can only have the price change if `showVariantImage` is enabled. Is that right? My point here is that `showVariantImage` doesn't reflect the fact that the price could change.

docs/pages/search-ui/results.mdx
```diff
@@ -86,12 +86,51 @@ function Example() {
...
+        }),
+        variables: new Variables({
+          resultsPerPage: 6,
+          q: 'macbook',
+        }),
+      }}
+      searchOnLoad
+    >
+      <Results showVariantImage />
+    </SearchProvider>
```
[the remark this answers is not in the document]
tuanddd: I tried using Unsplash's random url but it will also be a random image but repeated 5 times.

zlatanpham (approved):

PR #534, opened 2021-07-05
fix: filter empty label in list filter

tuanddd:


packages/search-ui/src/Filter/ListFilter.tsx
```diff
@@ -86,24 +95,29 @@ const ListFilter = (props: Omit<ListFilterProps, 'type'>) => {
   }, [JSON.stringify(selected)]);
 
   const Control = multi ? Checkbox : Radio;
-  const filtered = searchable ? options.filter((o) => o.label.toLowerCase().includes(query.toLowerCase())) : options;
-  const slice = filtered.length > limit;
+  const filteredEmptyLabel = options.filter(
+    (o) => !isNullOrUndefined(o.label) && isString(o.label) && !isEmpty(o.label),
+  );
```
zlatanpham, line 100: Should we use `useMemo` here? as the list doesn't need to be updated after a query filter.

packages/search-ui/src/Filter/ListFilter.tsx
```diff
@@ -86,24 +95,29 @@ const ListFilter = (props: Omit<ListFilterProps, 'type'>) => {
...
+  const filteredQuery = searchable
```
zlatanpham, line 101: Naming it `filteredQuery` doesn't sound right to me. The name implies a string variable rather than a collection.

packages/search-ui/src/Filter/ListFilter.tsx
```diff
@@ -86,24 +95,29 @@ const ListFilter = (props: Omit<ListFilterProps, 'type'>) => {
   }, [JSON.stringify(selected)]);
 
   const Control = multi ? Checkbox : Radio;
-  const filtered = searchable ? options.filter((o) => o.label.toLowerCase().includes(query.toLowerCase())) : options;
-  const slice = filtered.length > limit;
+  const filteredEmptyLabel = options.filter(
+    (o) => !isNullOrUndefined(o.label) && isString(o.label) && !isEmpty(o.label),
```
zlatanpham, line 99: I'm wondering if we need to use `isString` here. Even if a label is a number, is it fine to keep it?

tuanddd: Done

packages/search-ui/src/Filter/ListFilter.tsx
tuanddd: Done

packages/search-ui/src/Filter/ListFilter.tsx
tuanddd: Done

zlatanpham (approved):

PR #557, opened 2021-07-14
feat: add internationalization

tuanddd:


.changeset/four-starfishes-grin.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sajari/react-search-ui': minor
+---
+
+Add support for German, Dutch, Spanish, French and Vietnamese
```
zlatanpham, line 5:
```suggestion
Add support for German, Dutch, Spanish and French.
```

packages/search-ui/src/i18n/locale/nl.ts
```diff
@@ -0,0 +1,65 @@
...
+  },
+  sorting: {
+    label: 'Soort',
+    mostRelevantOption: 'Meest relevant',
+  },
+  summary: {
+    results: '{{count}} {{object}} voor "{{query}}"',
+    noQueryResults: '{{count}} {{object}}',
+    latency: '({{time}} secs)',
+    alternative: 'Search instead for',
```
zlatanpham, line 58: Looks like we are still missing some translations?

tuanddd: Yes, there are still some missing translations.

packages/search-ui/src/i18n/locale/de.ts
```diff
@@ -0,0 +1,65 @@
+export default {
+  common: {
```
zlatanpham, line 2: Maybe add an interface for the translation JSON? as it could help us avoid missing a translation. Also, It will be helpful as later we want to export the settings so users can input their own translations.

zlatanpham (approved):

PR #558, opened 2021-07-15
fix: list filter reset button does not show correctly

zlatanpham:
Fix list filter reset button doesn't show correctly. The bug can be found here - https://react.docs.sajari.com/search-ui/filter#list


wwalser (approved):
I'm approving assuming my understanding of the code is correct.

The initial filter may contain some selected items, but we only want to show if if the user has added additional options to the filter. Because of this, the old `selected.length > 0` would show the filter reset button even when only the default filters were in use. Now we check against `initial` to be sure that the user has actually made changes.

Assuming that's accurate, this looks good to go.

PR #559, opened 2021-07-15
fix: minimum character flicker

tuanddd:


packages/components/src/Combobox/index.tsx
```diff
@@ -51,13 +51,30 @@ const Combobox = React.forwardRef(function ComboboxInner<T>(props: ComboboxProps
...
+  const [items, setItems] = useState(itemsProp);
+  const lastItems = useRef(items);
+
+  useEffect(() => {
+    // Prevent flash of old dropdown items
+    if (value.length >= minimumCharacters && arraysEqual(items, []) && !arraysEqual(lastItems.current, itemsProp)) {
+      setItems(itemsProp);
+    }
+    lastItems.current = itemsProp;
+  }, [itemsProp]);
```
zlatanpham, line 71: Should we also include `minimumCharacters` in the array dependencies?

```suggestion
  }, [itemsProp, minimumCharacters]);
```

zlatanpham (commented):
Sorry for not being clear in our discussion but I meant we should maintain an internal state for items in `Input` rather than adding the state to`Combobox`.

packages/search-ui/src/Input/index.tsx
```diff
@@ -31,9 +31,10 @@ const Input = React.forwardRef((props: InputProps<any>, ref: React.Ref<HTMLInput
   const { search: searchAutocompleteFunc, completion, suggestions } = useAutocomplete();
   const { customClassNames, disableDefaultStyles = false, tracking } = useSearchUIContext();
   const { query } = useQuery();
-  // Use in the minimumCharacters implementation
-  const lastValue = useRef('');
-  let items: Array<any> = [];
+  const [internalSuggestions, setInternalSuggestions] = useState<Array<any>>([]);
+  const [items, setItems] = useState(internalSuggestions);
+  const lastItems = useRef(items);
+  const [internalValue, setInternalValue] = useState(query);
```
zlatanpham, line 37: Why do we need those states?

tuanddd: To be able to watch for changes and set the `items` state properly to pass down to combobox

zlatanpham: If it's because `query` is not a React state, maybe try to bring the check inside `onChange` or subscribe a change from `variables` so we don't have to maintain those extra states in the future.

packages/search-ui/src/Input/index.tsx
```diff
@@ -48,32 +48,41 @@ const Input = React.forwardRef((props: InputProps<any>, ref: React.Ref<HTMLInput
...
+            description,
+            image: isArray(image) ? image[0] : image,
+          };
+        }),
+      );
+    }
+  }, [mode, suggestions, maxSuggestions, results, useClickTracking, tracking]);
+
+  useEffect(() => {
+    if (!arraysEqual(lastItems.current, internalSuggestions) || mode === 'results') {
```
zlatanpham, line 79: Why do we need to check `mode === 'results'`?

packages/search-ui/src/Input/index.tsx
```diff
@@ -85,6 +94,10 @@ const Input = React.forwardRef((props: InputProps<any>, ref: React.Ref<HTMLInput
         onChange(value);
       }
 
+      if (value.length < minimumCharacters) {
+        setItems([]);
+      }
```
zlatanpham, line 99: Should we have the logic for `onVoiceInput`?

packages/search-ui/src/Input/index.tsx
```diff
@@ -31,9 +31,9 @@ const Input = React.forwardRef((props: InputProps<any>, ref: React.Ref<HTMLInput
...
   const { customClassNames, disableDefaultStyles = false, tracking } = useSearchUIContext();
   const { query } = useQuery();
-  // Use in the minimumCharacters implementation
-  const lastValue = useRef('');
-  let items: Array<any> = [];
+  const [internalSuggestions, setInternalSuggestions] = useState<Array<any>>([]);
+  const [items, setItems] = useState(internalSuggestions);
+  const lastItems = useRef(items);
 
   const search = useCallback(
```
zlatanpham, line 38: We recently added the `resetFilters` function so it still runs a new search if users previously selected a filter. I think we need to move the reset filter code inside the `minimumCharacters` check.

```jsx
const search = useCallback(
    (value: string) => {
      if (value.length >= minimumCharacters) {
        if (!retainFilters) {
          resetFilters();
        }
        searchFunc(value);
      }
    },
    [searchFunc, minimumCharacters],
  );
```

packages/search-ui/src/Input/index.tsx
tuanddd: Actually in `results` mode the two items array would never be the same everytime since they are list of objects so we can drop the mode check 👍

zlatanpham (approved):

PR #563, opened 2021-07-20
fix: results are not updated after a re-render

zlatanpham:
Fix the result image disappears after a re-render. It will happen if the response doesn't include a unique id for each entry.

### DIFF
--- a/packages/search-ui/src/Results/index.tsx
+++ b/packages/search-ui/src/Results/index.tsx
@@ -92,7 +92,7 @@ const Results = (props: ResultsProps) => {
           onClick={handleResultClicked}
           token={token}
           // eslint-disable-next-line no-underscore-dangle
-          key={values._id ?? i}
+          key={values._id ?? values.url ?? i}
           values={values}
           appearance={appearance}
           forceImage={hasImages}
--- a/.changeset/dry-flowers-raise.md
+++ b/.changeset/dry-flowers-raise.md
@@ -1,0 +1,6 @@
+---
+'sajari-sdk-docs': patch
+'@sajari/react-search-ui': patch
+---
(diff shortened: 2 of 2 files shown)


packages/search-ui/src/Results/index.tsx
```diff
@@ -92,7 +92,7 @@ const Results = (props: ResultsProps) => {
           onClick={handleResultClicked}
           token={token}
           // eslint-disable-next-line no-underscore-dangle
-          key={values._id ?? i}
+          key={values._id ?? values.url ?? i}
```
zlatanpham, line 95: Adding the guard but it could be possible for the response to not have `url` too. I think if the response always changes (it is), the results should be always fresh so we can force it to re-render if the results is updated.

```tsx
<React.Fragment key={results}>
   {results.map(...)}
</ReactFragment>
```

wwalser:
Feel free to message me in Slack about this. Am I correct in understanding that we're getting results back from the engine where each result is expected to have an `_id` but sometimes doesn't?

packages/search-ui/src/Results/index.tsx
wwalser: So wait… the result object id is sometimes missing?

zlatanpham:
> Feel free to message me in Slack about this. Am I correct in understanding that we're getting results back from the engine where each result is expected to have an `_id` but sometimes doesn't?
> 
> I ask because this sounds like a bug in the engine which would be worth fixing.

@wwalser your intuition is right. We could fix that by including `fields:''` in the body request so the response will come with `_id`. But I will wait for the confirmation from Dave if we should do it in the frontend.

PR #576, opened 2021-07-24
chore: cleanup typescript errors

wwalser:
Before change, tsc results in 626 errors.
After change, 18 errors.

It's possible to get this down to ~6 but that requires bumping and reconfiguring emotion. I've already spent more time on this than I wanted so I'm moving forward without doing that.


zlatanpham (approved):
Thanks for making the change  👍

PR #581, opened 2021-08-02
feat: support custom clicktracking url so we can test in non-prod environments

wwalser:
Gives the ability pass a custom click token url for use in staging environments. Allows end to end testing of clicktracking.

### DIFF
--- a/packages/hooks/src/ContextProvider/controllers/Pipeline.ts
+++ b/packages/hooks/src/ContextProvider/controllers/Pipeline.ts
@@ -13,14 +13,18 @@ const events = [EVENT_SEARCH_SENT, EVENT_RESPONSE_UPDATED, EVENT_RESULT_CLICKED]
 
 type QueryPipeline = ReturnType<Client['pipeline']>;
 
+type PipelineConfig = {
+  account: string;
+  collection: string;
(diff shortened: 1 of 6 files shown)


packages/hooks/src/ContextProvider/controllers/Pipeline.ts
```diff
@@ -69,11 +67,19 @@ export class Pipeline {
       p.version = name.version;
     }
 
+    const clientConfig: { [key: string]: string } = {};
+    if (typeof config.userAgent !== 'undefined') {
```
tuanddd, line 71: I think you can use `isNullOrUndefined` function from the `utils` package.

```suggestion
    if (!isNullOrUndefined(config.userAgent)) {
```

packages/hooks/src/ContextProvider/controllers/Pipeline.ts
```diff
@@ -69,11 +67,19 @@ export class Pipeline {
...
+      clientConfig.userAgent = config.userAgent;
+    }
+    if (typeof config.clickTokenURL !== 'undefined') {
```
tuanddd, line 74: Same here.

PR #586, opened 2021-08-04
feat: overlay z index and focus

wwalser:
feat: overlay z index and focus


packages/components/src/Modal/styles.ts
```diff
@@ -44,6 +44,14 @@ const animateModalOut = keyframes`
     }
 `;
 
+const zIndexOrDefault = (index?: number) => {
+  return index
+    ? css`
+        z-index: ${index};
```
tuanddd, line 50: We should use the `isNumber` check to make sure that passing `0` doesn't return `z-max`

zlatanpham (commented):
Once Tuan's comment is resolved, I think it's all good to merge.

tuanddd (approved):

PR #597, opened 2021-08-18
feat: add support for features of Result component to result template

tuanddd:
When changing to use result template, we need to keep the following existing features that is offered out-of-the-box when using the `Result` component:
- [x] Ability to render statuses (new arrival, on sale, out of stock) -> extracted to `useProductStatuses` common hook
- [x] Ability to render price (sale price, original price) -> extracted to `useRenderPrice` common hook
- [x] Internationalization -> pass the localized text to the result template for further use
- [x] Ability to change image on hover -> The plan is to use DOM manipulation to add hover logic to an `img` element (perhaps use something like `data-sj-second-image` attribute to hold the second image url)
- [x] Ability to show product variant images -> DOM manipulation
- [x] ClickTracking
- [x] PosNegTracking

### DIFF
--- a/.changeset/rotten-candles-attack.md
+++ b/.changeset/rotten-candles-attack.md
@@ -1,0 +1,6 @@
+---
+'@sajari/react-search-ui': minor
+'@sajari/react-sdk-utils': minor
+---
+
+Add support for features of Result component to result template
--- a/.changeset/rotten-candles-attack.md
+++ b/.changeset/rotten-candles-attack.md
@@ -1,6 +1,7 @@
 ---
 '@sajari/react-search-ui': minor
 '@sajari/react-sdk-utils': minor
+'@sajari/react-components': patch
 ---
 
 Add support for features of Result component to result template
--- a/packages/search-ui/src/Results/index.tsx
+++ b/packages/search-ui/src/Results/index.tsx
@@ -108,6 +108,7 @@ const Results = (props: ResultsProps) => {
         resetKeys={[`${resultTemplate.html}${resultTemplate.css}`]}
       >
         <TemplateResults
+          showVariantImage={rest.showVariantImage}
           results={results.map((r) => r.values)}
           resultTemplate={resultTemplate}
           resultContainerTemplateElement={resultContainerTemplateElement}
--- a/packages/search-ui/src/Results/index.tsx
+++ b/packages/search-ui/src/Results/index.tsx
@@ -6,6 +6,7 @@ import { ErrorBoundary } from 'react-error-boundary';
 import { useTranslation } from 'react-i18next';
 
 import { useSearchUIContext } from '../ContextProvider';
+import { useClickTracking } from '../hooks';
 import mapResultFields from '../utils/mapResultFields';
 import { checkValidResultTemplate } from './checkValidResultTemplate';
 import Message from './components/Message';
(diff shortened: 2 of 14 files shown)


tuanddd:
Ready for review.

packages/search-ui/src/Results/index.tsx
```diff
@@ -16,10 +17,36 @@ import { ResultsProps, ResultValues } from './types';
...
+  const untrackedResults = React.useMemo(
+    () => (rawResults ? mapResultFields<ResultValues>(rawResults, fields) : undefined),
+    [rawResults],
+  );
+  const { disableDefaultStyles = false, customClassNames, viewType, setViewType, tracking } = useSearchUIContext();
+  const { handleResultClicked, posNegLocalStorageManager } = useTracking();
+  const results = React.useMemo(() => {
+    return untrackedResults?.map((data) => {
+      const { values, token } = data;
+      const { href, onClick: clickTrackingOnClick } = useClickTracking({
```
wwalser, line 29: These calls to useClickTracking and usePosNegTracking are violations of the [Rules of Hooks](https://reactjs.org/docs/hooks-rules.html) and can cause problems if subsequent authors fail to understand how they are used.

There are two options for cases like this:
1. If these hooks don't actually contain the use of any built-in react hooks such as useMemo or useState, then they can simply be renamed so that it's obvious that they are not _actually_ hooks.
2. If they are actually hooks and call order matters, then we need to find a way to move them to the top level of some React component. The simplest way would be to write a wrapper component for Result and TemplateResult that calls these for each result items.

My reading of useClickTracking and usePosNegTracking is that they aren't actually hooks (they don't use any built-in React hooks and call order doesn't matter) so we can simply rename the functions to make it clear that they are simple data manipulation utilities.

packages/search-ui/src/Results/index.tsx
```diff
@@ -16,10 +17,36 @@ import { ResultsProps, ResultValues } from './types';
...
-  ]);
-  const { disableDefaultStyles = false, customClassNames, viewType, setViewType } = useSearchUIContext();
+  const untrackedResults = React.useMemo(
+    () => (rawResults ? mapResultFields<ResultValues>(rawResults, fields) : undefined),
+    [rawResults],
+  );
+  const { disableDefaultStyles = false, customClassNames, viewType, setViewType, tracking } = useSearchUIContext();
+  const { handleResultClicked, posNegLocalStorageManager } = useTracking();
+  const results = React.useMemo(() => {
+    return untrackedResults?.map((data) => {
```
wwalser, line 27: This introduced another loop over the results. Perhaps it would be better to have this tracking data management inside both <Result and <TemplateResult ?

Given that we paginate an additional loop isn't the end of the world but given the note below, I do wonder if this extra loop is an antipattern.

wwalser (changes requested):
Unfortunately I've found an architectural note that I think is substantial.

Sorry about the additional effort, but I think we should consider moving the tracking data manipulation back into Result and duplicate it inside of TemplateResult. I'm happy to hear your ideas.

tuanddd:
Sorry accidentally hit the request review button, this is not ready yet.

.changeset/rotten-candles-attack.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sajari/react-search-ui': minor
+'@sajari/react-sdk-utils': minor
```
wwalser, line 3: utils is used in most of the other libraries so you may need to bump those as well in order to get the changes distributed everywhere?

packages/search-ui/src/Results/useHoverImage.ts
```diff
@@ -0,0 +1,87 @@
...
+import { useCallback, useEffect, useState } from 'react';
+
+import { TemplateResultProps } from './components/TemplateResult/types';
+import { ResultValues } from './types';
+
+interface Props extends Pick<TemplateResultProps, 'showVariantImage'> {
+  image: ResultValues['image'];
+}
+
+export function useHoverImage(props: Props) {
```
wwalser, line 11: Mind explicitly defining the export type so that it's clear to readers without having to inspect with typescript or read the code?

wwalser (approved):
Looks good. You may want additional version bumps in the changeset but that's up to you. One other comment that is up to your discretion as well.

zlatanpham (approved):

PR #598, opened 2021-08-22
feat: replace tailwind's config units to EM

huygn:


packages/utils/src/styles/tailwind-unit-replace.ts
```diff
@@ -0,0 +1,64 @@
...
+        val = replacer(value);
+        break;
+    }
+    replacedTheme[key] = val;
+  }
+
+  return replacedTheme;
+}
+
+export const replaceTailwindUnit = (opts: Options) => (userConfig: TailwindConfig): TailwindConfig => {
```
zlatanpham, line 36: Do you think adding a unit test for this function would be useful to be better understanding how it works?

zlatanpham (commented):
LGTM. As long as you can fix the failing tests, it's all good to merge.

zlatanpham (dismissed):
LGTM 👍 . Can you also add a changeset file by running `yarn changeset` on your local before we merge this?

zlatanpham (approved):

zlatanpham:
@wwalser the code looks good to me. Please take a look if you have any suggestions, otherwise, I think we can ship it.

wwalser:
My hope for this change is that we would not introduce significant mutations at runtime. 
That said, I'm keen enough to get this shipped that we should move forward as-is.

wwalser (approved):

huygn:
> My hope for this change is that we would not introduce significant mutations at runtime.
> That said, I'm keen enough to get this shipped that we should move forward as-is.

This one runs in build time, `twin.macro` (and `tailwindcss`) reads it for configuration, the content of this file won't be included in the final build

wwalser:
Oh, I misunderstood how this was included into adjacent packages. Nice!
Thanks for clarifying.

PR #616, opened 2021-09-02
feat: shadow root support

huygn:
related to https://github.com/sajari/search-widgets/pull/132/#pullrequestreview-742904608, which caused by `Downshift`'s default to using `window` to execute `add/removeEventListener`. While they support `environment` prop to support other contexts, we also need to follow to allow passing custom context & support shadow DOM integration.

Open question: I used `shadowRoot` prop but `environment` prop seems to support `iframe` as well, so how about we just pass-thru `environment` prop from `SearchProvider`? (which seems more straight forward to understand from a user perspective)


packages/search-ui/src/ContextProvider/types.ts
```diff
@@ -13,6 +13,8 @@ export interface SearchUIContextProviderValues {
   /** View mode of the results */
   viewType?: ResultViewType;
   setViewType: (type: ResultViewType) => void;
+  /** shadow root support */
+  shadowRoot?: ShadowRoot | null;
```
zlatanpham, line 17: > Open question: I used `shadowRoot` prop but `environment` prop seems to support `iframe` as well, so how about we just pass-thru `environment` prop from `SearchProvider`? (which seems more straight forward to understand from a user perspective)

@huygn in that case, we just need to rename the prop to `environment` and enable it to accept `HTMLIFrameElement` as well, right?

zlatanpham (dismissed):
LGTM. Can you add a changeset file and include the release for `components` and `search-ui`?

packages/search-ui/src/ContextProvider/types.ts
huygn: yes and no, we can just use `Environment` prop provided by downshift

huygn: actually, it'd be `downshiftEnvironment` prop at `SearchUIContextProviderValues`

zlatanpham: @huygn I'm good with the change if you can do it.

packages/search-ui/src/ContextProvider/types.ts
```diff
@@ -13,6 +14,8 @@ export interface SearchUIContextProviderValues {
   /** View mode of the results */
   viewType?: ResultViewType;
   setViewType: (type: ResultViewType) => void;
+  /** allow passing custom environment for downshift */
```
zlatanpham, line 17:
```suggestion
  /** Allow passing custom environment for downshift in case the UI is rendered within a different window context - https://github.com/downshift-js/downshift#environment */
```

zlatanpham (approved):

PR #636, opened 2021-09-08
fix: clear container when toggling show variant image option

tuanddd:
Need to clear the images container when toggling the variant image option or the old images will stack up.

### DIFF
--- a/packages/search-ui/src/Results/useProductImages.ts
+++ b/packages/search-ui/src/Results/useProductImages.ts
@@ -148,10 +148,14 @@ export function useProductImages(props: Props): UseProductImagesOutput {
 
   useEffect(() => {
     const { image } = values;
-    if (container && showVariantImage && isArray(image)) {
-      const style = document.createElement('style');
-      style.id = 'sj-result-template-default-style';
-      style.textContent = `
+    if (container) {
+      while (container.firstChild) {
+        container.removeChild(container.firstChild);
+      }
+      if (showVariantImage && isArray(image)) {
+        const style = document.createElement('style');
+        style.id = 'sj-result-template-default-style';
+        style.textContent = `
         [data-sj-variant-image]::before {
           padding-bottom: calc(100%);
           content: "";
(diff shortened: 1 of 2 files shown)


packages/search-ui/src/Results/useProductImages.ts
```diff
@@ -148,24 +148,29 @@ export function useProductImages(props: Props): UseProductImagesOutput {
...
   useEffect(() => {
     const { image } = values;
-    if (container && showVariantImage && isArray(image)) {
-      const style = document.createElement('style');
-      style.id = 'sj-result-template-default-style';
-      style.textContent = `
+    if (container) {
+      while (container.firstChild) {
+        container.removeChild(container.firstChild);
+      }
```
zlatanpham, line 154: The use of `while` look looks strange. Seems to me it's an alternative for an if statement?

packages/search-ui/src/Results/useProductImages.ts
```diff
@@ -148,24 +148,29 @@ export function useProductImages(props: Props): UseProductImagesOutput {
...
-      images.forEach((img) => {
-        container.appendChild(img);
-      });
-    } else {
-      document.querySelector('#sj-result-template-default-style')?.remove();
+        document.head.appendChild(style);
+        const images = image.map(getCreateImageElementFunc(setActiveImageIndex));
+        images.forEach((img) => {
+          container.appendChild(img);
+        });
```
zlatanpham, line 170: Consider using [DocumentFragment](https://developer.mozilla.org/en-US/docs/Web/API/DocumentFragment) for the sake of performance.

packages/search-ui/src/Results/useProductImages.ts
tuanddd: No this is intentional, this will remove all child nodes of `container` until `container` is empty.

zlatanpham: I don't know that can do the trick. I was assuming that we need to reassign the container at the end of the statement.

zlatanpham: Ah no, see that. Looks good though!

zlatanpham (approved):

PR #638, opened 2021-09-09
fix: render price logic when showing variant images

tuanddd:
This PR will add another variable called `isCurrentOnSale` this is different from `isOnSale` from that `isOnSale` will be true if at least one of the variant products is on sale. On the other hand, `isCurrentOnSale` is only of that variant product.


zlatanpham (approved):
The code looks good to me, however, the calculation is getting complex that I think we might want to cover them with tests soon.

PR #670, opened 2021-09-28
fix: filter process too many items

tuanddd:


tuanddd:
The reason behind the slowness is the frontend filtering and sorting a huge amount of options that are returned back from backend (up to ~6k items in some cases) so we have to compromise by limiting the size of options. This obviously will remove some data that is being requested but in exchange for a more responsive UI

wwalser:
I'm going to take some time to investigate this a bit more. Eliminating data from the set doesn't seem like the correct solution. Looking at the code, it appears we loop over the same data set around 12 times in useFilter. My intuition is that looping over 6k items shouldn't take very long.

Do you have a data set handy where this reproduces easily?

zlatanpham:
Closing in favor of https://github.com/sajari/sdk-react/pull/674

PR #691, opened 2021-11-04
feat: debounce search backspace

tuanddd:


packages/search-ui/src/Input/index.tsx
```diff
@@ -35,17 +35,25 @@ const Input = React.forwardRef((props: InputProps<any>, ref: React.ForwardedRef<
...
+        if (lastValue.current.length > value.length) {
+          searchDebounce(value);
+        } else {
+          searchFunc(value);
+        }
+        lastValue.current = value;
       }
     },
-    [searchFunc, minimumCharacters, retainFilters, resetFilters],
+    [searchFunc, lastValue.current, minimumCharacters, retainFilters, resetFilters],
```
zlatanpham, line 56: I don't think the change of a value from `useRef` could revalidate a `useCallback`. So including `lastValue.current` in the dependency array doesn't make sense here.

packages/search-ui/src/Input/index.tsx
```diff
@@ -46,14 +46,27 @@ const Input = React.forwardRef((props: InputProps<any>, ref: React.ForwardedRef<
...
   const searchValue = useCallback(
     (value: string) => {
       if (value.length >= minimumCharacters) {
         if (!retainFilters) {
           resetFilters();
         }
-        searchFunc(value);
+        // cannot use useDebounce because we want to always clear timeout but not always trigger the debounce function
+        // this is to fix the bug where deleting multiple characters and then immediately start typing and the input value got magically removed because the remove debounce function is still running
+        window.clearTimeout(searchDebounceRef.current);
```
zlatanpham, line 60: Maybe it's safe to clear the timeout in unmount phase?

zlatanpham (approved):
LGTM. One question.

zlatanpham (approved):

PR #695, opened 2021-11-09
Don't send unchanged range filter values

wwalser:
fix: avoid sending range filter when min and max values are set to the min and max of the collection


zlatanpham (approved):
LGTM 👍

PR #700, opened 2021-11-25
feat: redirects are now implemented in search-ui package

wwalser:
feat: redirects are now implemented in search-ui package

Won't push this out until mid next week to avoid problems with major shopping days in some locales.

### DIFF
--- a/packages/search-ui/src/Input/index.tsx
+++ b/packages/search-ui/src/Input/index.tsx
@@ -1,7 +1,7 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 import { Combobox } from '@sajari/react-components';
 import { useAutocomplete, useQuery, useSearchContext } from '@sajari/react-hooks';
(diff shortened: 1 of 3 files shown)


packages/search-ui/src/Input/index.tsx
```diff
@@ -29,14 +30,24 @@ const Input = React.forwardRef((props: InputProps<any>, ref: React.ForwardedRef<
...
-  const { search: searchAutocompleteFunc, completion, suggestions } = useAutocomplete();
+  const {
+    search: searchAutocompleteFunc,
+    completion,
+    suggestions,
+    redirects,
+    searching: autocompleteSearching,
+  } = useAutocomplete();
+  const redirectsRef = useRef(redirects);
+  redirectsRef.current = redirects;
```
tuanddd, line 41: Unless I'm missing something or this line is already covered by the above line? 🤔

tuanddd (approved):

PR #712, opened 2021-12-13
feat: add pinned item identifier to search-ui result component

jkaho:


wwalser (approved):

wwalser:
Assuming you've tested this and it works, the code look very simple and ready to go.

wwalser:
I comment before merging it purely because it's in a draft state. Not sure if you intended that or not.

jkaho:
@wwalser I did intend to put it in a draft state because I wasn't totally sure if I was going about it the right way. If you don't see any obvious issues with this change though, I think it should be safe to merge.

PR #714, opened 2021-12-16
rebrand: change sajari to search.io in react docs

wwalser:
rebrand: change sajari to search.io in react docs


wwalser:
Not switching the repo names over yet. Not sure when we're going to get to that.

tuanddd (dismissed):

PR #724, opened 2022-01-04
fix: enter submit input component

tuanddd:


packages/search-ui/src/Input/index.tsx
```diff
@@ -167,31 +167,33 @@ const Input = React.forwardRef((props: InputProps<any>, ref: React.ForwardedRef<
   const onKeyDownMemoized = useCallback(
     (e: React.KeyboardEvent<HTMLInputElement>) => {
       const { value } = e.currentTarget;
-      if (e.key === 'Enter' && (mode === 'typeahead' || mode === 'suggestions' || mode === 'standard')) {
-        if (!retainFilters) {
-          resetFilters();
+      if (e.key === 'Enter') {
```
zlatanpham, line 170: There is one problem that holding `Enter` will trigger the search to run non-stop. Doesn't feel really nice to me and I'm seeing other folks sort it out by blurring the input.

packages/search-ui/src/Input/index.tsx
```diff
@@ -167,31 +167,35 @@ const Input = React.forwardRef((props: InputProps<any>, ref: React.ForwardedRef<
...
```
zlatanpham, line 170: @tuanddd if the mode is instant, it's pretty useless to have the enter key trigger. Normally, the results come right away but pressing `Enter` unnecessarily triggers the same search.

zlatanpham (approved):

PR #734, opened 2022-01-31
chore: upgrade typescript and nextjs

wwalser:
chore: upgrade typescript and nextjs


packages/components/src/types/speech-recognition.ts
```diff
@@ -0,0 +1,104 @@
+export type SpeechRecognitionErrorCode =
```
JasonBerry, line 1: Should we use [@types/web](https://www.npmjs.com/package/@types/web) instead of copy-pasting our own copy?

JasonBerry (approved):
Just the one comment about the speech recognition types but that's it.

packages/components/src/types/speech-recognition.ts
wwalser: Tried that, types/web doesn't contain the speech recognition type anymore as it's only available in some browsers. If you read enough comments in typescript PRs and issues it becomes clear that they are open to people adding it to definitely typed but not to adding it back to types/web for the moment.

JasonBerry: Ah, didn't realise it had been removed.

PR #743, opened 2022-02-17
fix: should trigger search on Enter instant Input

tuanddd:


wwalser (dismissed):

tuanddd:
@wwalser this PR was a result of me using the shopify UI builder in the `console` and the search bar (`instant` mode) won't make a search if there are less than 3 characters.

Thanh thought it was unnecessary, what about yours?

wwalser (approved):

PR #746, opened 2022-02-17
fix: render order of banners

zlatanpham:
Change how banners are rendered to follow the index of the list of banners and results instead of the index of the results. For example, given a list of:
- Banners: `[{position: 1}, {position: 2}]`
- Results: `[A, B]`

**Before**
Output: `[{position:1}, A, {position: 2}, B]`

**After**
Output: `[{position:1}, {position: 2}, A, B]`

This change is to ensure we cover the case if users want to place 2 banners next to each other.


tuanddd:
LGTM

wwalser:
Algorithmically, the same comments apply here as in my console review.

wwalser (approved):

PR #750, opened 2022-02-23
fix: add resize observer for custom template

tuanddd:
This PR wrap the custom result template container with ResizeObserver (similar to the normal `Result` component). The reason is to get rid of extra responsive CSS that we have to inject along with the code snippet and to provide responsive support for banners.

### DIFF
--- a/.changeset/late-actors-destroy.md
+++ b/.changeset/late-actors-destroy.md
@@ -1,0 +1,5 @@
+---
+'@sajari/react-search-ui': patch
+---
+
+Add ResizeObserver for custom templates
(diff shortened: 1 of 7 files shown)


wwalser (dismissed):

zlatanpham:
Note that it always displays one column when I pull down the code to do the testing. Also, if this requires the code snippet in the console to change then it can unexpectedly break the UI of existing customers if they did use a result template. I think it is something we should put in consideration.

tuanddd:
It shouldn't be, I'll have a look

zlatanpham (approved):
LGTM

wwalser (approved):

PR #752, opened 2022-02-28
fix: banner should not expand row height

tuanddd:
This should fix the issue where the banner height is larger than the row's items height and therefore expanding the row height (see below)
Screen Shot 2022-02-28 at 11 41 50

After the fix it should look like this
Screen Shot 2022-02-28 at 11 31 01

However, when the banner's height is smaller than the item's height then the following happens (the banner doesn't span all the full height), maybe @zlatanpham or @chidojiro would have an idea.
Screen Shot 2022-02-28 at 11 42 57

### DIFF
--- a/.changeset/hot-files-hang.md
+++ b/.changeset/hot-files-hang.md
@@ -1,0 +1,5 @@
+---
+'@sajari/react-search-ui': patch
+---
+
+Banner should not increase the row height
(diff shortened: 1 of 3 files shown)


zlatanpham:
I'm currently looking into the issue

wwalser (dismissed):

zlatanpham:
Fixed the issue mentioned in the description:

https://user-images.githubusercontent.com/12707960/157042072-f87e817b-2715-4179-9a5d-a2d416c7e69a.mov

wwalser (dismissed):

zlatanpham (approved):

PR #753, opened 2022-03-02
feat: integrate SearchIOAnalytics for simplified event tracking

JasonBerry:


packages/hooks/src/ContextProvider/controllers/Response.ts
```diff
@@ -109,6 +112,13 @@ export class Response {
     return this.response !== undefined ? (this.response.get('time') as number) : undefined;
   }
 
+  /**
+   * Return queryId from the response.
+   */
+  public getQueryId(): string | undefined {
+    return this.response !== undefined ? (this.response.get('queryId') as string) : undefined;
```
wwalser, line 119: I get the impression that `this.response?.get('queryId') as string | undefined` would be sufficient here? Using a `Map` instead of a typed `{}` for `response` was a design mistake. Means you have to constantly type things. I wonder if we can roll out of that without impacting anyone.

packages/search-ui/src/Input/index.test.tsx
```diff
@@ -0,0 +1,92 @@
+import 'whatwg-fetch';
```
wwalser, line 1: 🙏

packages/search-ui/src/Input/index.tsx
```diff
@@ -93,19 +94,20 @@ const Input = React.forwardRef((props: InputProps<any>, ref: React.ForwardedRef<
         results.splice(0, maxSuggestions).map((result) => {
           const { values, token } = result;
           const { description, image, title } = values;
-          const { href, onClick } = applyClickTracking({ tracking, values, token });
+          const { href, onClick: clickTrackingOnClick } = applyClickTracking({ tracking, values, token });
+          const { onClick: eventTrackingOnClick } = applyEventTracking({ tracking, values, searchIOAnalytics });
```
wwalser, line 98: Maybe there's an opportunity here for a hook, specific to the Input component as it doesn't need to be a part of the SDK, which performs both of these and picks between the two. I'm a little weary to have both inline like this.

Imagine the file is ~10 lines long and 3 of them are a comment explaining why it's necessary to do both.

packages/search-ui/src/utils/applyClickTracking.ts
```diff
@@ -1,11 +1,11 @@
-import { ClickTracking, PosNegTracking, ResultClickedFn, Token } from '@sajari/react-hooks';
+import { ClickTracking, EventTracking, PosNegTracking, ResultClickedFn, Token } from '@sajari/react-hooks';
 import { isEmpty, isFunction, isString, isValidURL, noop } from '@sajari/react-sdk-utils';
 
 import { ResultValues } from '../Results/types';
 
 export interface ApplyClickTrackingParams {
   token: Token | undefined;
-  tracking: ClickTracking | PosNegTracking | undefined;
+  tracking?: ClickTracking | PosNegTracking | EventTracking;
```
wwalser, line 8: I may have other nits and questions but architecturally, this is my one concern.

I wonder if it's wise to have this be yet another type that tracking could be instead of a wholely new field whose type is absolutely certain. To me, tracking is something that users of the API were meant to choose between. New simplified tracking is not another choice among those but _the_ way to go moving forward.

The number of times that we need to use `instanceof` is a smell IMO.

wwalser (approved):
I'm okay if 'this is the way we should go' but I did flag one bigger thing which I'm genuinely concerned about and would like to hear your thoughts on. Happy to meet or type it out.

Thanks for gardening as you go 👍.

packages/hooks/src/ContextProvider/controllers/Response.ts
JasonBerry: `this.response?.get('queryId') as string | undefined` would indeed be sufficient, I was simply following the established pattern but I might clean all of these up while I'm at it to make them consistent.

Agreed that the `Map` adds unnecessary verbosity with very little benefit, and in fact works against the ability to type the response nicely and means we need to keep casting all the time. I'd like to re-work the `Response` object in a separate release rather than dropping it into this one.

packages/search-ui/src/utils/applyClickTracking.ts
JasonBerry: Agreed. Will refactor the tracking work next week (existing and new) to not need any of this.

packages/search-ui/src/Input/index.tsx
JasonBerry: I think the refactor next week will address this.

packages/hooks/src/useTracking/index.ts
```diff
@@ -8,11 +6,9 @@ function useTracking(): UseTrackingResult {
     resultClicked,
     search: { pipeline },
   } = useContext();
-  const client = pipeline.getClient();
   return {
-    consumeInteractionToken: client.interactionConsume,
+    consumeInteractionToken: pipeline.getClient().interactionConsume,
     handleResultClicked: resultClicked,
-    posNegLocalStorageManager: new PosNegLocalStorageManager(client),
```
JasonBerry, line 15: I've removed this from the hook, does that require a major release you think?

wwalser (dismissed):
Fantastic. Great encapsulation with onResultClicked.

packages/hooks/src/ContextProvider/controllers/Pipeline.test.ts
```diff
@@ -56,4 +56,25 @@ describe('Pipeline', () => {
 
     expect(pipeline.getClient().config.userAgent).toEqual(userAgent);
   });
+
+  it('should call tracking.bootstrap on constructor', () => {
```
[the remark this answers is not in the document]
JasonBerry: That's a good call. There's a few I'll update with this in mind.

packages/hooks/src/useTracking/index.ts
wwalser: Technically it should, but I really don't want to.

Does keeping it around cause any knock on issues? Perhaps a @todo to remove on next major release?

wwalser (approved):

PR #756, opened 2022-03-03
fix: remove unwanted resetFilters from autocomplete search

zlatanpham:
Remove `resetFilters` when calling autocomplete search. This trigger of the method is not only unnecessary, but it also introduces a mess in SearchContext to override the input value.

https://search-io.atlassian.net/browse/SF-705

### DIFF
--- a/.changeset/popular-clouds-learn.md
+++ b/.changeset/popular-clouds-learn.md
@@ -1,0 +1,5 @@
+---
+'@sajari/react-search-ui': patch
+---
+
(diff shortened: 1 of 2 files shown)


wwalser:
@tuanddd you introduced this line in https://github.com/sajari/sdk-react/commit/366b3de997b7854061b7d1be101298109ad7634e. Will this bring back the bug you were resolving in that change?

tuanddd:
No it should be good.

wwalser (approved):

PR #758, opened 2022-03-11
fix: performance issue on results

chidojiro:
resolve: https://search-io.atlassian.net/jira/software/projects/SF/boards/29?selectedIssue=SF-641

diagnose: rerenders happen so many times that makes the even loop busy and slow to process input events

solution: make items of Results independent from any context

### DIFF
--- a/packages/search-ui/src/Results/index.tsx
+++ b/packages/search-ui/src/Results/index.tsx
@@ -21,7 +21,16 @@ const Results = (props: ResultsProps) => {
   const results = React.useMemo(() => (rawResults ? mapResultFields<ResultValues>(rawResults, fields) : undefined), [
     rawResults,
   ]);
-  const { disableDefaultStyles = false, customClassNames, viewType, setViewType } = useSearchUIContext();
+  const {
+    disableDefaultStyles = false,
+    customClassNames,
+    viewType,
+    setViewType,
+    currency,
+    tracking,
+    ratingMax,
+    language,
+  } = useSearchUIContext();
   const { query } = useQuery();
   const {
     defaultAppearance,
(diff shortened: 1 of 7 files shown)


packages/hooks/CHANGELOG.md
```diff
@@ -199,15 +199,15 @@
   Before:
 
   ```js
-  pipeline.listen('result-clicked', url => {
+  pipeline.listen('result-clicked', (url) => {
```
zlatanpham, line 202: I know it's just Eslint stuff but this change is not supposed to be included in the PR.

zlatanpham (commented):
This code looks good to me but it's not quite clear how different the performance is after the change. Can we have a before/after comparison or performance profiling evidence for the update?

packages/search-ui/src/Results/index.tsx
```diff
@@ -49,6 +48,9 @@ const Results = (props: ResultsProps) => {
     results,
   ]);
 
+  const { currency, tracking, ratingMax, language } = useSearchUIContext();
```
wwalser, line 51: Is this the only substantial change, moving the context hook up the tree?

wwalser (changes requested):

wwalser (dismissed):

zlatanpham (approved):

zlatanpham:
One note is that we are using `JSON.stringify` comparison in [result component](https://github.com/sajari/sdk-react/blob/52ef27ef4c83d769ebc841a856591af887d4b140/packages/search-ui/src/Results/components/Result/index.tsx#L331). This check looks expensive and I'm wondering if we can have an equivalent check by `_id`?

zlatanpham:
> One note is that we are using `JSON.stringify` comparison in [result component](https://github.com/sajari/sdk-react/blob/52ef27ef4c83d769ebc841a856591af887d4b140/packages/search-ui/src/Results/components/Result/index.tsx#L331). This check looks expensive and I'm wondering if we can have an equivalent check by `_id`?

@wwalser just reminded me of one issue that makes my previous assumption incorrect:

> I think the problem with _id in sdk-react is that _id may not always be returned in search queries. It is returned by default, but customers can change the default result fields of their pipeline.

So we are good without touching the part for now.

wwalser:
Rebase on master then feel free to pull the trigger. Thanks @chidojiro

zlatanpham (approved):

PR #759, opened 2022-03-11
fix: incorrect french translation for error message

zlatanpham:


packages/search-ui/src/i18n/locale/fr.ts
```diff
@@ -17,13 +17,13 @@ const fr: Locale = {
...
     },
-    generic: "Quelque chose s'est mal passé. S'il vous plaît vérifier à nouveau.",
+    generic: 'Une erreur est survenue. Veuillez réessayer ultérieurement.',
   },
   filter: {
     all: 'Tout',
     rangeOver: 'Plus de {{value}}',
     rangeUnder: 'Sous {{value}}',
-    reset: 'Réinitialiser',
+    reset: 'Merci. Réessayer',
```
jkaho, line 26: `Merci. Réessayer` means `Thank you. Try again`. I think we should get rid of the `Merci.`

Maybe they meant to make it part of the error message? Even so I think it'd be okay to get rid of it entirely as I don't recall us using "thank you" in our English error messages.

Can wait for Wes's input though 🙂

zlatanpham: Updated. Your thought should be right.

JasonBerry (approved):

PR #766, opened 2022-03-16
fix: make sure pagination is memoized

chidojiro:
resolve: https://search-io.atlassian.net/jira/software/projects/SF/boards/29?selectedIssue=SF-641

changes: for some reasons, React.memo needs an explicit comparing callback to memoize correctly, which Pagination is missing => add a comparing callback to it

the only change is the addition of the callback, the rest is from auto formatting (on save)

### DIFF
--- a/packages/components/src/Pagination/index.tsx
+++ b/packages/components/src/Pagination/index.tsx
@@ -20,209 +20,212 @@ const defaultI18n = {
   current: 'Page {{page}}, current page',
 };
 
-const Pagination = React.memo((props: PaginationProps) => {
-  const { ref, width } = useResizeObserver<HTMLDivElement>();
-  const {
-    language,
-    totalResults,
-    resultsPerPage,
-    page,
-    pageCount,
-    onChange,
-    i18n: i18nProp,
-    buttonClassName,
-    activeClassName = '',
-    nextClassName,
-    prevClassName,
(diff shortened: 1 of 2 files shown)


packages/components/src/Pagination/index.tsx
```diff
@@ -20,209 +20,212 @@ const defaultI18n = {
...
+          key="next"
+          css={tw`m-0`}
+        >
+          &#8203;
+          <IconChevronRight />
+        </Button>
+      </ButtonGroup>
+    );
+  },
+  (prev, next) => JSON.stringify(prev) === JSON.stringify(next),
```
zlatanpham, line 227: It really concerns me that we are using  `JSON.stringify` too often rather than using the traditional Referential Equality check. Maybe we should see if we can use `useCallback` or `useMemo` for the outer props instead?

wwalser:
I'm not convinced that this component is slow. Do we have perf data showing that Pagination is consuming meaningful cycles on render?

packages/components/src/Pagination/index.tsx
chidojiro: I don't know whether the issue comes from preact or not, but it needs a deep compare to do a memo

chidojiro: I could be wrong, let me check that again

PR #770, opened 2022-03-22
fix: enter behavior in input

tuanddd:
This PR removes the blur behavior after the user presses enter because it doesn't allow them to search again immediately. Instead, a `readonly` attribute is added to preserve the focus while still preventing the user from repeatedly spamming enter key, after a search is done the attribute is removed.

Credit goes to @JasonBerry for the solution idea.

### DIFF
--- a/packages/search-ui/src/Input/index.tsx
+++ b/packages/search-ui/src/Input/index.tsx
@@ -162,12 +162,13 @@ const Input = React.forwardRef((props: InputProps<any>, ref: React.ForwardedRef<
 
   const onKeyDownMemoized = useCallback(
     (e: React.KeyboardEvent<HTMLInputElement>) => {
+      const input = e.target as HTMLInputElement;
       const { value } = e.currentTarget;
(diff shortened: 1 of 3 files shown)


packages/search-ui/src/Input/index.tsx
```diff
@@ -167,7 +167,7 @@ const Input = React.forwardRef((props: InputProps<any>, ref: React.ForwardedRef<
         const closestForm = (e.target as HTMLInputElement).closest('form');
         // Blur input to avoid Enter spamming, but only if the input is not wrapped by any form, otherwise the submit event won't be triggered
```
JasonBerry, line 168: We're not blurring any more, so update this comment.

Also, `readonly` doesn't prevent enter spamming, but you can do a check to see if `(e.target as HTMLInputElement).readOnly` is true and skip and of the processing.

Please add a test for this behaviour.  😊

JasonBerry (changes requested):

JasonBerry (approved):

PR #771, opened 2022-03-29
fix: product price null as return value

tuanddd:
## Problem
Currently, the `useRenderPrice` hook return the value `null` in some cases, this is not a problem with our standard `Result` component because we already perform checks before rendering. However, this will potentially break in result template mode because users wouldn't expect `productPrice` to be null (maybe it's not too obvious/not documented somewhere)

## The fix
Instead of returning null in those cases, we return the same object shape but with an empty string, this will have the same effect as if the user would have to write a check before rendering in the template code themselves

### DIFF
--- a/packages/search-ui/src/Results/useRenderPrice.ts
+++ b/packages/search-ui/src/Results/useRenderPrice.ts
@@ -1,4 +1,4 @@
-import { formatPrice, isArray, isEmpty } from '@sajari/react-sdk-utils';
+import { formatPrice, isArray, isEmpty, isNullOrUndefined } from '@sajari/react-sdk-utils';
 
 import type { ResultValues } from './types';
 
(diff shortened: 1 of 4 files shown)


wwalser (dismissed):

packages/search-ui/src/Results/useRenderPrice.ts
```diff
@@ -35,7 +48,11 @@ export function useRenderPrice({
     dataOriginalPrice = originalPriceProp.slice(1);
   }
   const activePrice = isArray(price) ? price[activeImageIndex] ?? price : price;
-  if (isEmpty(activePrice)) return null;
+  if (isEmpty(activePrice)) {
+    return {
+      displayPrice: '',
```
chidojiro, line 53: this is repeated. I think it should be extracted to a variable like `defaultPrice`

wwalser (approved):

PR #774, opened 2022-03-31
fix: banner absolute position

tuanddd:
## Problem
The banner image container is having absolute position style so when there is only a banner on a row it will collapse to zero height because no other result item is there to establish the height.

## Solution
Only apply the absolute style when the column span of that banner is larger than the total number of columns.


wwalser (approved):

wwalser:
SF-760

PR #778, opened 2022-04-20
fix: add Tab key as another way to complete suggestion

tuanddd:
## Problem
- Using Tab key doesn't expand/complete the suggestion in typeahead mode of Combobox

## Solution
Add the Tab key

## Checklist
- [x] Implementation
- [x] Tests

SF-578

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,6 +72,7 @@
     "eslint-plugin-simple-import-sort": "^7.0.0",
     "husky": "^4.3.8",
     "jest": "^26.6.3",
+    "jest-environment-jsdom-sixteen": "^2.0.0",
     "jest-svg-transformer": "^1.0.0",
     "lerna": "^3.22.1",
     "lint-staged": "^10.5.4",
--- a/package.json
+++ b/package.json
@@ -72,7 +72,6 @@
     "eslint-plugin-simple-import-sort": "^7.0.0",
     "husky": "^4.3.8",
     "jest": "^26.6.3",
-    "jest-environment-jsdom-sixteen": "^2.0.0",
     "jest-svg-transformer": "^1.0.0",
     "lerna": "^3.22.1",
     "lint-staged": "^10.5.4",
--- a/packages/components/src/Combobox/test/Combobox.test.tsx
+++ b/packages/components/src/Combobox/test/Combobox.test.tsx
@@ -41,8 +41,10 @@ describe('Combobox', () => {
   });
 
   it('Resume default behavior of Tab key after completing the suggestion', async () => {
-    // so we don't get `getSelection is not defined` error
-    document.getSelection = () => null;
+    if (!document.getSelection) {
+      // so we don't get `getSelection is not defined` error
+      document.getSelection = () => null;
+    }
     const { getByTestId } = render(<Combobox mode="typeahead" completion="apple mac studio" />);
     const element = getByTestId('combobox-input') as HTMLInputElement;
     element.focus();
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -72,6 +72,7 @@
     "eslint-plugin-simple-import-sort": "^7.0.0",
     "husky": "^4.3.8",
     "jest": "^26.6.3",
+    "jest-environment-jsdom-sixteen": "^2.0.0",
```
zlatanpham, line 75: I wonder what was the reason we need to add this dependency.

zlatanpham (commented):
There is a similar UX for the suggestions mode as when you hit `tab`, it will auto pick the first item from the dropdown as the selected value. But I'm open if you want to do it in a different PR.

image

package.json
JasonBerry: `jest-environment-jsdom-sixteen` is deprecated, we shouldn't add it.

tuanddd: This line in the test
https://github.com/sajari/sdk-react/pull/778/files#diff-a1cc8e7375dcbeb36ae4103d4e85572ebf3d95fb1193cfd572ac38d0b4e19030R41

will throw an error, something like `...getSelection() is not a function`, in order to make it work the jest jsdom-sixteen env must be used.

tuanddd: Ah saw the deprecation message, maybe we should upgrade Jest instead?

JasonBerry: Yep, upgrade Jest

tuanddd: I ended up not upgrading Jest, because of two reasons:
1. It breaks the whole `test` pipeline since it is a breaking change (26 -> 27)
2. I think the cause wasn't really originated from Jest but rather testing-library/user-event because I tried downgrading it to v13 and the error message changed, but again I couldn't risk downgrading for just a test case

As a result, I had to manually mock the `getSelection` function so that it won't yell anymore.

zlatanpham (approved):

package.json
JasonBerry: > It breaks the whole test pipeline since it is a breaking change (26 -> 27)

What was it that broke? Checking the [changelog](https://github.com/facebook/jest/blob/main/CHANGELOG.md#2700) it looks as though the breaking changes should be minimal and relatively easy to resolve. Can do that in another PR anyway though, so don't worry about it for now!

> As a result, I had to manually mock the `getSelection` function so that it won't yell anymore.

Best to mock that out in a way that won't potentially break when we upgrade to a version of jsdom that contains the missing method, e.g.:
```js
if (!document.getSelection) {
  document.getSelection = () => null;
}
```

packages/components/src/Combobox/test/Combobox.test.tsx
```diff
@@ -9,4 +12,42 @@ describe('Combobox', () => {
...
+    const element = getByTestId('combobox-input') as HTMLInputElement;
+    element.focus();
+    await userEvent.keyboard('{ArrowRight}');
+
+    expect(element.value).toBe('');
+  });
+
+  it('Resume default behavior of Tab key after completing the suggestion', async () => {
+    // so we don't get `getSelection is not defined` error
+    document.getSelection = () => null;
```
JasonBerry, line 45: Mentioned it in a different comment, but best to check for the existence of `document.getSelection` before potentially overwriting it.

JasonBerry (dismissed):

wwalser (dismissed):

JasonBerry (approved):

zlatanpham (approved):

PR #784, opened 2022-04-22
feat: add ability to pass custom SearchIOAnalytics instance to EventTracking

JasonBerry:
This should enable easier ad-hoc event tracking support, like:
```typescript
import { EventTracking, SearchIOAnalytics } from '@sajari/react-hooks';

const account = '1594153711901724220';
const collection = 'bestbuy';
const searchIOAnalytics = new SearchIOAnalytics(account, collection);
const eventTracking = new EventTracking('id', undefined, searchIOAnalytics);
const pipeline = new Pipeline({ account, collection, }, 'query', eventTracking);

// e.g. in add to cart buttons - doesn't require React context, etc
searchIOAnalytics.track('add_to_cart', product.id);
```

In this way the React components share the same `SearchIOAnalytics` instance as wherever else we might need to track events, so the latest `queryId` will be updated by the pipeline query and correctly bound to the events tracked elsewhere.

### DIFF
--- a/.changeset/two-actors-exist.md
+++ b/.changeset/two-actors-exist.md
@@ -1,0 +1,5 @@
+---
+'@sajari/react-hooks': minor
+---
+
+feat: add ability to pass custom SearchIOAnalytics instance to EventTracking
--- a/packages/hooks/src/ContextProvider/controllers/tracking/EventTracking.test.ts
+++ b/packages/hooks/src/ContextProvider/controllers/tracking/EventTracking.test.ts
@@ -1,4 +1,4 @@
-import { Client, TrackingType } from '@sajari/sdk-js';
+import { Client, SearchIOAnalytics, TrackingType } from '@sajari/sdk-js';
 
 import { EVENT_TRACKING_RESET } from '../../events';
 import { Response } from '../Response';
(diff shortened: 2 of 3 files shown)


tuanddd (commented):
LGTM

wwalser:
```
const searchIOAnalytics = new SearchIOAnalytics(account, collection);
const eventTracking = new EventTracking('id', undefined, searchIOAnalytics);
const pipeline = new Pipeline({ account, collection, }, 'query', eventTracking);

// e.g. in add to cart buttons - doesn't require React context, etc
searchIOAnalytics.track('add_to_cart', product.id);
```

Can `track()` be called safely with having been passed to an EventTracking constructor and that instance to a pipeline? If so, the example code feels fairly strange as it's not clear why `eventTracking` and `pipeline` are being instantiated.

I think most cases where we want to instantiate ad-hoc tracking, we won't also have search and results appearing on the same screen. Given that, sharing the same instance is probably not of paramount concern.

wwalser (approved):

JasonBerry:
> Can `track()` be called safely with having been passed to an EventTracking constructor and that instance to a pipeline? If so, the example code feels fairly strange as it's not clear why `eventTracking` and `pipeline` are being instantiated. I get from the text that follows that it may be important for instance re-use across other components but I think most cases where we want ad-hoc tracking, we won't also have search results appearing on the same screen. Given that, sharing the same instance is probably not of paramount concern.

It was a contrived example purely to illustrate how it could be used without needing to pull the `SearchIOAnalytics` instance out of the pipeline. What's more likely is that you'd construct a pipeline for use in the search box in a website/app and you share the `SearchIOAnalytics` instance between that and the rest of the page where you're tracking ad-hoc events. Perhaps you'd throw it on the `window` for easy access.

You'd really want to share the same instance on the page because `SearchIOAnalytics` only pulls the existing event data out of `localStorage` when it's instantiated.

wwalser:
Great, thanks for the future information.

PR #785, opened 2022-04-25
fix: unstable input test

tuanddd:
## Problem
One of the Input test that tests for the `readonly` attribute being set properly when the user spams the Enter key is unstable, it fails occasionally.

## Solution
Remove the statement where it is failing, I think it's reasonable given that we don't actually want to test for the inside mechanism of how it works but rather than just the expected result that we will get.

### DIFF
--- a/.changeset/fast-ducks-tan.md
+++ b/.changeset/fast-ducks-tan.md
@@ -1,5 +1,0 @@
----
-'@sajari/react-search-ui': patch
----
-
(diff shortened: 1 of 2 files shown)


zlatanpham (approved):

JasonBerry (changes requested):
Changeset not required. We don't need a version number bump for a fixed test.

JasonBerry (approved):

PR #792, opened 2022-05-09
fix: add data-testid for range filter track

chidojiro:
## Changes
Add data-tesid for range filter input so that it can be queried for testing


.changeset/ninety-avocados-join.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sajari/react-components': patch
+---
+
+add data-testid for range filter track
```
zlatanpham, line 5: Should also upgrade search-ui package

zlatanpham (changes requested):

zlatanpham (approved):

PR #794, opened 2022-05-11
feat: add sync URL state functionality

zlatanpham:
Motivation: It’s super useful to have the search UI state synchronizing with the browser URL and we had the functionality internally developed in the search widgets. However, when users build the search UI with React SDK (or JS SDK), they have to develop the feature by themselves and the work could be challenging even for an experienced developer. Thus, the goal of the change is to have the sync URLs functionality as a generic solution in the React SDK so it can be easily used by both our internal development (search-widgets, demo,…) and the outside world.

### API proposal
To enable the synchronize the state of the search UI with the browser URL, set the prop `syncURLState` for `SearchProvider`. It will work out of the box with any UI that consumes data from our hook components.

```jsx
import { SearchProvider } from '@sajari/react-hooks';

<SearchProvider {...} syncURLState>
  <YourSearchUI />
</SearchProvider>
```

By default, it supports state sync for the core params including `q`, `filters`, `sort` and `resultsPerPage` but we can extend other options via `extendedParams` field:

```jsx
import { SearchProvider } from '@sajari/react-hooks';

const defaultViewType = 'list';
const [viewType, setViewType] = useState(defaultViewType);

<SearchProvider
     syncURLState={
       extendedParams: [
          {
            key: 'viewType',
            value: viewType,
            defaultValue: defaultViewType,
            callback: setViewType,
          }
        ]
     }
>
  <YourSearchUI />
</SearchProvider>
```


packages/hooks/src/useQueryParams/index.ts
```diff
@@ -0,0 +1,67 @@
+import { isSSR } from '@sajari/react-sdk-utils';
+import { useCallback, useEffect, useState } from 'react';
+
+import useDebounce from '../useDebounce';
+import { ParamValue, UseQueryParamParams } from './types';
+
+function useQueryParams(key: string, options?: UseQueryParamParams) {
+  const { callback, defaultValue, debounce, replace = false } = options ?? {};
+  const [internalValue, setInternalValue] = useState<ParamValue | undefined>(undefined);
+  const debouncedValue = debounce ? useDebounce(internalValue, debounce) : internalValue;
```
chidojiro, line 10: I think it's better not using a conditional hook

packages/hooks/src/useQueryParams/index.ts
```diff
@@ -0,0 +1,67 @@
+import { isSSR } from '@sajari/react-sdk-utils';
+import { useCallback, useEffect, useState } from 'react';
+
+import useDebounce from '../useDebounce';
+import { ParamValue, UseQueryParamParams } from './types';
+
+function useQueryParams(key: string, options?: UseQueryParamParams) {
```
chidojiro, line 7: is it correct it's supposed for one key-value param only?
then I think the name should be singular

packages/hooks/src/useQueryParams/index.ts
```diff
@@ -0,0 +1,67 @@
...
+  const { callback, defaultValue, debounce, replace = false } = options ?? {};
+  const [internalValue, setInternalValue] = useState<ParamValue | undefined>(undefined);
+  const debouncedValue = debounce ? useDebounce(internalValue, debounce) : internalValue;
+  const getValue = () => (!isSSR() ? new URLSearchParams(window.location.search).get(key) ?? '' : '');
```
chidojiro, line 11: (opinionated) (!isSSR() && new URLSearchParams(window.location.search).get(key)) || '' might be more readable

packages/hooks/src/useQueryParams/index.ts
```diff
@@ -0,0 +1,67 @@
...
+    const url = new URL(window.location.href);
+    const prevSearch = url.search;
+
+    if ((!Array.isArray(val) && val && val !== defaultValue) || (Array.isArray(val) && val.length > 0)) {
+      url.searchParams.set(key, val.toString());
+    } else if (val?.toString().length === 0 || val === defaultValue || (Array.isArray(val) && val.length === 0)) {
+      url.searchParams.delete(key);
+    }
+
+    if (prevSearch === url.search) {
```
chidojiro, line 27: should this be place right after prevSearch declaration?

zlatanpham: I think no. The value of `url.search` could be modified from the previous `if-else` block.

packages/hooks/src/useQueryParams/index.ts
zlatanpham: The suggestion is shorter to write but the old code still looks more readable to me so I think we should keep it as it was.

packages/hooks/src/useQueryParams/index.ts
zlatanpham: Updated

zlatanpham:
Change the status back to WIP as it's waiting for https://github.com/sajari/sdk-react/pull/798. This might require an API update as well.

wwalser:
👀

packages/hooks/src/URLStateSync/index.tsx
```diff
@@ -0,0 +1,167 @@
...
+      allowSetParam.current = true;
+    }
+  }, [response]);
+
+  useEffect(() => {
+    if (results && filter.getFrozen()) {
+      // wait for the cycle of React to end then releasing the frozen state
+      setTimeout(() => {
+        filter.setFrozen(false);
+      });
```
wwalser, line 95: With react embracing async more and more, little hacks like this are prone to breaking.
Did you check to see if useEffect without the setTimeout is sufficient?

Another solution is, instead of relying on useEffect to determine the initial render of the component would be to make result aware of the fact that it's the initial result set.

wwalser (approved):

zlatanpham:
@wwalser I've made a couple of changes since your last review in https://github.com/sajari/sdk-react/pull/794/commits/59e2794c9bf0d4c28518cbd2a35c9a34d831f6e6

wwalser (dismissed):

packages/hooks/src/URLStateSync/index.tsx
zlatanpham: No, it doesn't work without `setTimeout`. I believe it's the trickiest part of the SDK to make the range filter sync URL work properly. I will try to look for a better solution in another PR.

PR #807, opened 2022-05-25
Migrate yarn

wwalser:
chore: implement steps necessary to move local commands and build system to modern yarn


packages/hooks/src/useTracking/index.test.tsx
```diff
@@ -1,9 +1,9 @@
-import { FieldDictionary, Pipeline } from '@sajari/react-hooks';
+// import { FieldDictionary, Pipeline } from '@sajari/react-hooks';
```
JasonBerry, line 1: Remove

JasonBerry (approved):
When we update yarn next time is there a simple command that will update the files we've committed to our repo under `.yarn` or is that all pretty manual?

wwalser:
> When we update yarn next time is there a simple command that will update the files we've committed to our repo under .yarn or is that all pretty manual?

`yarn set version *whatever*`

wwalser:
Builds and tests are running now but I have a single test failure that I can't reproduce locally.

wwalser:
@zlatanpham I've hit my head on this for a while, can you investigate why that test is failing? 
Also happy for @tuanddd to have a look as I think he's the original author of the test. I'd be curious to know if it's failed in this manner elsewhere.

JasonBerry:
We're still using node 12.x in the release.yml (instead of the version we specify in nvm) and run-test.yml doesn't even specify a version… who knows what it's using. That could be a potential source of the failure.

wwalser:
Yup, I'll switch to using the nvmrc file.

wwalser:
-_-

.github/workflows/release.yml
```diff
@@ -20,13 +20,31 @@ jobs:
           fetch-depth: 0
           persist-credentials: false
 
-      - name: Use Node.js 12.x
-        uses: actions/setup-node@v1
+      - name: Setup node
+        uses: actions/setup-node@v2
         with:
-          version: 12.x
+          node-version-file: '.nvmrc'
```
JasonBerry, line 26: TIL: `node-version-file` is a thing. I now feel the need to fix all the actions into which I introduced that bash script to manually use nvm.

JasonBerry (approved):

PR #818, opened 2022-05-31
feat: add page param to StateSyncURL component

tuanddd:
## What this PR does
- Add the page param to the list of watchers

https://search-io.atlassian.net/browse/SF-865


zlatanpham (commented):
The code looks good to me but it seems the search API is called twice. Can you check that?

image

PR #820, opened 2022-06-07
feat: update PoweredBy text

jkaho:
Change `PoweredBy` component text from 'Powered by Search.io' to 'Site search by Search.io`.

[SF-866](https://search-io.atlassian.net/browse/SF-866)
_____

### Mono
Screen Shot 2022-06-07 at 4 32 54 pm
### Color
Screen Shot 2022-06-07 at 4 37 03 pm

### DIFF
--- a/packages/components/src/PoweredBy/index.tsx
+++ b/packages/components/src/PoweredBy/index.tsx
@@ -20,7 +20,7 @@ const PoweredBy = React.forwardRef((props: PoweredByProps, ref?: React.Ref<HTMLA
       {...props}
     >
       <Box as="span" css={styles.label}>
-        Powered by
+        Site search by
       </Box>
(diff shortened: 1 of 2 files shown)


packages/components/src/PoweredBy/index.tsx
```diff
@@ -20,7 +20,7 @@ const PoweredBy = React.forwardRef((props: PoweredByProps, ref?: React.Ref<HTMLA
       {...props}
     >
       <Box as="span" css={styles.label}>
-        Powered by
+        Site search by
```
jkaho, line 23: @wwalser do you think it's worth creating a new component called `SiteSearchBy` and deprecating `PoweredBy`, or would just changing the text suffice?

wwalser: I think just the text change is sufficient.

wwalser (approved):

PR #823, opened 2022-06-22
fix: url param injection due to changing default resultsPerPage value

zlatanpham:
Fix the issue when changing the default value of `resultsPerPage` to a value that is different from `15` will cause URL param injection.


wwalser (approved):
Happy to pull the trigger on this for now.

I do wonder if there's a more general solution along the lines of 'only write to URL if state deviates from initialization defaults'.

PR #856, opened 2022-09-05
Tailwind upgrade

JasonBerry:
* Minimum React version is now 17 to keep in step with minimum specified by Sajari UI & Next – identified some potential event cleanup timing issues and rectified those
* Custom, pseudo-element ring styles replaced with native CSS focus ring styling (less buggy and more performant)


.changeset/afraid-boxes-invent.md
```diff
@@ -0,0 +1,10 @@
+---
+'sajari-sdk-docs': major
+'@sajari/react-components': major
+'@sajari/react-hooks': major
+'@sajari/react-search-ui': major
+'@sajari/server': major
+'@sajari/react-sdk-utils': major
```
wwalser, line 7: What is utils used by? Asking because I notice that Tailwind is a direct dependency there instead of a peer dependency.

wwalser (dismissed):
Looks good. Should tailwind be made a peer dependency in utils?

jkaho (approved):
A small thing, probably doesn't matter, but noting that the invalid Select button focus outline isn't red.
Screen Shot 2022-09-05 at 4 36 12 pm

Invalid Checkbox:
Screen Shot 2022-09-05 at 4 42 40 pm

JasonBerry:
> A small thing, probably doesn't matter, but noting that the invalid Select button focus outline isn't red. Screen Shot 2022-09-05 at 4 36 12 pm
> 
> Invalid Checkbox: Screen Shot 2022-09-05 at 4 42 40 pm

I just kept it the same as it is currently: https://react.docs.search.io/components/select

.changeset/afraid-boxes-invent.md
JasonBerry: `utils` is used by `components`, `hooks`, `search-ui` and `server`. All of the Tailwind bits are added as dependencies in that package (as well as a common Tailwind config) and the other packages use Tailwind transitively as a result. We'd be unable to build without Tailwind as a dependency due to how we use Emotion and Twin, so I think moving it to peer would be the wrong call.

wwalser (approved):

PR #858, opened 2022-11-08
fix: upgrade downshift and remove breaking changes

beatrice94:
Upgraded Downshift from 6.1.0 -> 7.0.1
Removed references to GetComboboxProps

### DIFF
--- a/.changeset/healthy-books-own.md
+++ b/.changeset/healthy-books-own.md
@@ -1,0 +1,5 @@
+---
+'@sajari/react-components': major
+---
+
+upgraded downshift to v7.0.1
--- a/.changeset/healthy-books-own.md
+++ b/.changeset/healthy-books-own.md
@@ -1,5 +1,0 @@
----
-'@sajari/react-components': major
----
-
-upgraded downshift to v7.0.1
--- a/package.json
+++ b/package.json
@@ -93,8 +93,5 @@
     "@types/react-dom": "^17.0.11",
     "postcss": "^8.4.16"
   },
-  "packageManager": "yarn@3.2.1",
-  "dependencies": {
-    "yalc": "^1.0.0-pre.53"
-  }
+  "packageManager": "yarn@3.2.1"
 }
--- a/package.json
+++ b/package.json
@@ -93,5 +93,8 @@
     "@types/react-dom": "^17.0.11",
     "postcss": "^8.4.16"
   },
-  "packageManager": "yarn@3.2.1"
+  "packageManager": "yarn@3.2.1",
+  "dependencies": {
+    "yalc": "^1.0.0-pre.53"
+  }
 }
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -45,8 +45,7 @@
     "react-focus-lock": "^2.5.0",
     "react-ranger": "^2.1.0",
     "smoothscroll-polyfill": "^0.4.4",
-    "use-resize-observer": "^7.0.0",
-    "yalc": "^1.0.0-pre.53"
+    "use-resize-observer": "^7.0.0"
   },
   "devDependencies": {
     "babel-jest": "^26.6.3",
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -41,11 +41,12 @@
     "body-scroll-lock": "^3.1.5",
     "classnames": "^2.2.6",
     "color": "^3.1.3",
-    "downshift": "^6.1.0",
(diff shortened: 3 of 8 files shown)


package.json
```diff
@@ -93,5 +93,8 @@
     "@types/react-dom": "^17.0.11",
     "postcss": "^8.4.16"
   },
-  "packageManager": "yarn@3.2.1"
+  "packageManager": "yarn@3.2.1",
+  "dependencies": {
+    "yalc": "^1.0.0-pre.53"
+  }
```
jkaho, line 99: I think you can remove this `yalc` dependency. I assume it was added when you were trying to install it globally
```suggestion
```

packages/components/package.json
```diff
@@ -41,11 +41,12 @@
...
     "classnames": "^2.2.6",
     "color": "^3.1.3",
-    "downshift": "^6.1.0",
+    "downshift": "7.0.1",
     "react-focus-lock": "^2.5.0",
     "react-ranger": "^2.1.0",
     "smoothscroll-polyfill": "^0.4.4",
-    "use-resize-observer": "^7.0.0"
+    "use-resize-observer": "^7.0.0",
+    "yalc": "^1.0.0-pre.53"
```
jkaho, line 49: You can remove this too
```suggestion
```

packages/components/src/Select/index.tsx
```diff
@@ -43,8 +43,8 @@ const Select = React.forwardRef((props: SelectProps, ref?: React.Ref<HTMLDivElem
     const { changes, type } = actionAndChanges;
 
     switch (type) {
-      case useSelect.stateChangeTypes.MenuKeyDownEnter:
-      case useSelect.stateChangeTypes.MenuKeyDownSpaceButton:
+      case useSelect.stateChangeTypes.ToggleButtonKeyDownEnter:
```
jkaho, line 46: What does this change do?

yarn.lock
```diff
@@ -8441,13 +8442,6 @@ __metadata:
   languageName: node
   linkType: hard
 
-"compute-scroll-into-view@npm:^1.0.16":
```
jkaho, line 8444: Was this supposed to be deleted?

jkaho (changes requested):

packages/components/src/Select/index.tsx
beatrice94: This change is because the name of the properties were changed. This can be found in the migration documents [here](https://github.com/downshift-js/downshift/blob/master/src/hooks/MIGRATION_V7.md)

yarn.lock
beatrice94: Oh this was no meant to be deleted, will update

beatrice94: It looks like this was replaced when running yarn

jkaho: As discussed, this is fine

beatrice94 (commented):
Made requested changes to package files

jkaho:
It looks like the size and position of the combobox text has changed.

**Before**
Screen Shot 2022-11-09 at 5 06 57 pm

**After**
Screen Shot 2022-11-09 at 5 06 52 pm

Hopefully it's an easy fix!

wwalser (approved):
The styling change is a good catch. Once that's fixed, feel free to pull the trigger with this.

Nice upgrade @beatrice94

wwalser:
I've changed some configuration on the Netlify builds. I believe those should pass once you push another commit.

packages/components/src/Combobox/styles.ts
```diff
@@ -27,7 +27,7 @@ export function getInputSpacingStyles(size?: ComboboxSize) {
 
     case 'md':
     default:
-      return tw`pl-9`;
+      return tw`pl-10 text-base`;
```
jkaho, line 30: I think leaving the padding at 9 should be fine
```suggestion
      return tw`pl-9 text-base`;
```

Screen Shot 2022-11-14 at 3 20 52 pm

jkaho (changes requested):
Small nit!

.changeset/healthy-books-own.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sajari/react-components': major
```
jkaho, line 2: I've checked in with Wes, who thinks this should be a minor bump

jkaho (changes requested):

jkaho (approved):

PR #863, opened 2022-11-30
fix: add check so page can never be 0

beatrice94:


wwalser (dismissed):

packages/hooks/src/ContextProvider/index.tsx
```diff
@@ -339,7 +339,7 @@ const ContextProvider: React.FC<SearchProviderValues> = ({
       const { pipeline, variables } = search as Required<ProviderPipelineConfig>;
       const { config } = searchState;
 
-      variables.set({ [config.pageParam]: String(page) });
```
JasonBerry, line 342: We should add a test for this change, but there isn't already existing tests for this component at all. 😢

PR #873, opened 2023-02-02
fix: enable non-case-sensitive redirects from the `Input` component

jkaho:
Currently, adding a `string-lowercase` step to the query/autocomplete pipeline enables the query `Tires` to return the redirect for `tires`, e.g.:
```
redirects: { 
    tires: { 
        id: "abc", 
        target: "https://foo.com",
        token: "123"
    } 
}
```
But the `react-search-ui` `Input` component will only look for redirects based on the raw query `Tires`, not the transformed version `tires`.

Prefer to use the (possibly transformed) query returned from the search response over the raw input query to look up redirects.

[DAU-249](https://algolia.atlassian.net/browse/DAU-249)

___

https://user-images.githubusercontent.com/72959522/216412720-8a8ee086-3656-4516-a60d-e979b2959389.mov

### DIFF
--- a/packages/hooks/src/useAutocomplete/index.ts
+++ b/packages/hooks/src/useAutocomplete/index.ts
@@ -1,11 +1,13 @@
 import { useContext } from '../ContextProvider';
 
 function useAutocomplete() {
+  // TODO: might not make sense to return the last response from this hook?
   const {
-    autocomplete: { suggestions, search, completion, searching, redirects },
+    autocomplete: { response, suggestions, search, completion, searching, redirects },
   } = useContext();
 
   return {
+    response,
     suggestions,
     search,
     completion,
--- a/packages/search-ui/src/Input/index.tsx
+++ b/packages/search-ui/src/Input/index.tsx
@@ -35,6 +35,7 @@ const Input = React.forwardRef((props: InputProps<any>, ref: React.ForwardedRef<
     suggestions,
     redirects,
     searching: autocompleteSearching,
+    response,
   } = useAutocomplete();
   const redirectsRef = useRef(redirects);
   redirectsRef.current = redirects;
@@ -47,6 +48,9 @@ const Input = React.forwardRef((props: InputProps<any>, ref: React.ForwardedRef<
   const inputRef = mergeRefs(ref, localRef);
   const lastValue = useRef(query);
   const searchDebounceRef = useRef(0);
+  const qOriginal = response?.getValues()?.get('q.original');
+  const responseQuery = useRef(qOriginal);
+  responseQuery.current = qOriginal;
 
   const searchValue = useCallback(
     (value: string, bypass = false) => {
@@ -174,7 +178,7 @@ const Input = React.forwardRef((props: InputProps<any>, ref: React.ForwardedRef<
           if (!retainFilters) {
             resetFilters();
           }
-          const redirectValue = redirectsRef.current[value];
+          const redirectValue = redirectsRef.current[responseQuery.current ?? value];
           if (!disableRedirects && redirectValue) {
             tracking.onRedirect(redirectValue);
             window.location.assign(redirectValue.token || redirectValue.target);
@@ -183,7 +187,7 @@ const Input = React.forwardRef((props: InputProps<any>, ref: React.ForwardedRef<
             // If we're performing an autocomplete search, wait a tick to recheck redirects before unloading
             e.preventDefault();
             setTimeout(() => {
-              const redirectTarget = redirectsRef.current[value];
+              const redirectTarget = redirectsRef.current[responseQuery.current ?? value];
               if (redirectTarget) {
                 tracking.onRedirect(redirectTarget);
                 window.location.assign(redirectTarget.token || redirectTarget.target);
--- a/.changeset/hip-carrots-count.md
+++ b/.changeset/hip-carrots-count.md
@@ -1,0 +1,6 @@
+---
+'@sajari/react-hooks': patch
+'@sajari/react-search-ui': patch
+---
+
+fix: enable non-case-sensitive redirects from the Input component


packages/hooks/src/useAutocomplete/index.ts
```diff
@@ -1,11 +1,13 @@
 import { useContext } from '../ContextProvider';
 
 function useAutocomplete() {
+  // TODO: might not make sense to return the last response from this hook?
```
jonathaningram, line 4: You're returning response because you need to access values (read: variables) in the calling code. Could we instead return variables from this? variables is a sibling of redirects, and we return redirects off this, so could we also not return variables?

Or if we need to distinguish these variables from the input ones maybe we need to name them `outputVariables` or `evaluatedVariables`

packages/search-ui/src/Input/index.tsx
```diff
@@ -174,7 +178,7 @@ const Input = React.forwardRef((props: InputProps<any>, ref: React.ForwardedRef<
           if (!retainFilters) {
             resetFilters();
           }
-          const redirectValue = redirectsRef.current[value];
+          const redirectValue = redirectsRef.current[responseQuery.current ?? value];
```
jonathaningram, line 181: So previously, if the user typed in `tir` (imagine they were intending to type in `tire`) and then they stopped, does the existing behavior mean that it was always trying to look up `redirects['tir']`? Which does seem odd to me.

packages/search-ui/src/Input/index.tsx
```diff
@@ -47,6 +48,9 @@ const Input = React.forwardRef((props: InputProps<any>, ref: React.ForwardedRef<
   const inputRef = mergeRefs(ref, localRef);
   const lastValue = useRef(query);
   const searchDebounceRef = useRef(0);
+  const qOriginal = response?.getValues()?.get('q.original');
```
jonathaningram, line 51: It's a bit of a magic string isn't it. Do we need to make this config or are there places we also bake in `q`?

jonathaningram (changes requested):
I suspect we're going to want to make a test, because it's not really clear to me that this change doesn't break subtle behavior we previously expected: We're changing the input change handler to not look at the changed `value` anymore, but instead use the evaluated/output variable.

Long shot, but we don't have any concept of feature flags in here do we? E.g. can't just put this behind a flag and test it out for a specific customer?

packages/search-ui/src/Input/index.tsx
jkaho: Yes, that's the behaviour I was seeing.

To clarify, what exactly doesn't seem right about that to you? I would think that, for the most part, we'd want to search for redirects that match the exact query string the user has inputted (with the exception of bypassing the whole case-sensitivity thing), no? What if there really was a redirect for `tir`?

packages/search-ui/src/Input/index.tsx
jkaho: The hooks package uses this default config internally:
```
export const defaultConfig: Config = {
  qParam: 'q',
  qOverrideParam: 'q.override',
  qSuggestionsParam: 'q.suggestions',
  resultsPerPageParam: 'resultsPerPage',
  pageParam: 'page',
  maxSuggestions: 10,
  fieldsParam: 'fields',
};
```
I could do something similar in the `search-ui` package for `qOriginalParam`

packages/search-ui/src/Input/index.tsx
jonathaningram: > What if there really was a redirect for tir?

Yep you're right. If the user types `tire` and there is a redirect for `tire`, then we should expect the redirect to trigger, rather than autocomplete kick it, correct it to, say, `tired` (which incidentally may also have a redirect of its own and it's unclear in that case who would win out of `tire` or `tired`). All that is to say that, looking at the value the user typed makes sense. Not sure why I thought it odd. 

I guess it's not clear to me if changing to look at `q.original` breaks stuff. Because the old behavior is such that on key down it uses the value in the input field and looks up redirects. And that value is available / changes on every on key down event. But the new behavior is such that it is going to ignore what the user typed, and look at `q.original` which is only available and only updated once a `Search` API call is done. And the time between the response to that API call and the user typing in the input is varied. And because of that, I'm not sure of the consequences of changing the behavior of this so (what appears to be) subtly.

packages/search-ui/src/Input/index.tsx
jonathaningram: Probably a good idea.

packages/search-ui/src/Input/index.tsx
jonathaningram: Just quickly, I'm not saying this change is wrong, I'm just unsure of the consequences :|

jkaho: Fair callouts 🙂 Better to be safe than sorry. I'll do more testing and look into writing a test, though I'm not sure rn how your concerns would be captured in said test

jkaho: Perhaps we could do one of those old env var flags we used to do in the console

packages/search-ui/src/Input/index.tsx
jkaho: I'm now realising that the `original` param of the `autocomplete` pipeline step might not be bound to a variable called `q.original`, it could be anything, so I may need to look at the actual pipeline to get the variable name

packages/hooks/src/useAutocomplete/index.ts
jkaho: There don't seem to be any `variables` returned in the search response, just `values` (which looks like it might be the equivalent of `variables` in the v4 search response?), e.g., this is what all the responses I'm getting look like (minus redirects for some):
```
{
    "values": {
        "q": "tires",
        "q.original": "tires"
    },
    "redirects": {
        "tires": {
            "id": "abc",
            "target": "https://www.performanceplustire.com/tires-for-sale/",
            "token": "123"
        }
    }
}
```
And the variables I think you're referring to are just default params that we set

```
{
    "q": "Tires",
    "resultsPerPage": 15,
    "fields": "",
}
```

packages/search-ui/src/Input/index.tsx
jkaho: > But the new behavior is such that it is going to ignore what the user typed, and look at q.original which is only available and only updated once a Search API call is done. And the time between the response to that API call and the user typing in the input is varied.

After thinking about this - the redirects we're referring to in this line
```
const redirectValue = redirectsRef.current[responseQuery.current ?? value];
```
come from the same search API call as the `q.original` value, no? Would this not mean that if we have a value for `redirectsRef` from the search response then we should also have one for `responseQuery`?

packages/hooks/src/useAutocomplete/index.ts
jonathaningram: Sorry `values` is the v2 term. We renamed to `variables` in v4. So they are one and the same.

packages/search-ui/src/Input/index.tsx
jonathaningram: If there are places in the library where we already ignore potentially renamed params then I think it's OK to keep doing that. Maybe a TODO note next to it saying that it won't work if the user has changed it in their pipeline.

packages/search-ui/src/Input/index.tsx
jonathaningram: > Perhaps we could do one of those old env var flags we used to do in the console

It would allow for quickly disabling it if we had to, but we'd still need to roll out the release and have it propagate to all end users. So there's still a chance of breaking users and having no fast rollback. Probably not going to be feasible to do any sort of feature flagging here.

jkaho:
Closing in favour of less bug-prone solution: https://github.com/sajari/sdk-react/pull/875

jkaho:
Closing in favour of less bug-prone solution: https://github.com/sajari/sdk-react/pull/875

PR #875, opened 2023-02-07
feat: make redirects via the input component case-insensitive

jkaho:
Assume users don't care about the case-sensitivity of redirects and lowercase all inputs when retrieving redirects, as well as the redirect keys themselves (the queries that trigger the redirects).

[DAU-249](https://algolia.atlassian.net/browse/DAU-249)

### DIFF
--- a/docs/pages/search-ui/input.mdx
+++ b/docs/pages/search-ui/input.mdx
@@ -133,6 +133,12 @@ function Example() {
 }
 ```
 
+### Note
+
+The `Input` component treats redirects as _case-insensitive_, meaning the user-inputted value `Computer` returns the redirect for the query `computer` and vice versa.
+
+The component does not support the handling of multiple redirects for queries of the same word with varied letter casing. For example, if you have created redirects for each of the queries `computer`, `Computer`, and `COMPUTER`, only one of them is used (dependent upon your browser).
+
 ## Results
 
 ```jsx
--- a/packages/search-ui/src/Input/index.test.tsx
+++ b/packages/search-ui/src/Input/index.test.tsx
@@ -101,6 +101,22 @@ describe('Input', () => {
     });
   });
 
+  it('ignores letter casing of input query on redirect lookup', async () => {
+    const onRedirectSpy = jest.spyOn(eventTrackingPipeline.getTracking(), 'onRedirect');
+    customRender(<Input data-testid="mysearch" mode="suggestions" />, {
+      search: { pipeline: eventTrackingPipeline },
+    });
+    const input = screen.getByTestId('mysearch');
+    input.focus(); // need this else we get TypeError: element.ownerDocument.getSelection is not a function
+    await user.keyboard('ShEeTs');
+    await waitFor(() => expect(input).toHaveTextContent('ShEeTs'));
+    await user.keyboard('{enter}');
+    expect(onRedirectSpy).toHaveBeenCalledWith({
+      ...redirectTarget,
+      token: `https://re.sajari.com/token/${redirectTarget.token}`, // sdk-js prepends the clickTokenURL
+    });
+  });
+
   it('calls onResultClick on results click', async () => {
     const onResultClickSpy = jest.spyOn(eventTrackingPipeline.getTracking(), 'onResultClick');
     customRender(<Input data-testid="mysearch" mode="results" />, {
--- a/packages/search-ui/src/Input/index.test.tsx
+++ b/packages/search-ui/src/Input/index.test.tsx
@@ -49,7 +49,7 @@ const server = setupServer(
         values: {
           q: 'sheets',
           'q.original': 'sheets',
-          'q.suggestions': 'sheets',
+          'q.suggestions': 'suggestion',
         },
         redirects: {
           sheets: redirectTarget,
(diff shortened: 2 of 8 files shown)


packages/search-ui/src/Input/index.tsx
```diff
@@ -174,7 +174,7 @@ const Input = React.forwardRef((props: InputProps<any>, ref: React.ForwardedRef<
           if (!retainFilters) {
             resetFilters();
           }
-          const redirectValue = redirectsRef.current[value];
+          const redirectValue = redirectsRef.current[value.toLowerCase()];
```
jonathaningram, line 177: I think this only works if the items in `redirectsRef.current` are also lowercased, no?

E.g. if the redirect is `Tires`, this never triggers if the user types `Tires` does it?

jonathaningram (changes requested):

packages/search-ui/src/Input/index.tsx
jkaho: Yeah, this assumes that their redirects are all lowercased. I would think we might want to build in some kind of case-sensitivity feature into the backend where the user can decide whether their redirect is case-sensitive or not upon creation? But yes you're right, if they have a redirect with capitals, then there will be no way to trigger it.

My concern with lowercasing the whole redirects object was that it could be a lot of work if the user has a lot of redirects, since we'd do it each time the search response is returned. But maybe it's safe to assume that they wouldn't have that many. WDYT?

jonathaningram: Yeah I don't think iterating over those objects in the response and transforming the keys to lowercase is that expensive.

jonathaningram: If we change the behavior such that looking up `Tires` in there no longer works, that breaks for every client using this which isn't already lowercasing. So we should ensure we lowercase the redirects coming from the backend response plus the input value.

jonathaningram (changes requested):
RCFV

packages/search-ui/src/Input/utils/index.test.ts
```diff
@@ -0,0 +1,19 @@
+import { lowercaseObjectKeys } from '.';
+
+test('lowercaseRedirects', () => {
+  const redirects = {
+    FoO: { id: '1', target: 'a' },
+    BAR: { id: '2', target: 'b' },
+    baz: { id: '3', target: 'c' },
+    Test: { id: '4', target: 'd' },
```
jonathaningram, line 8: Should you add a duplicate here e.g. `test` to prove that only one survives?

packages/search-ui/src/Input/index.test.tsx
```diff
@@ -49,7 +49,7 @@ const server = setupServer(
         values: {
           q: 'sheets',
           'q.original': 'sheets',
-          'q.suggestions': 'sheets',
+          'q.suggestions': 'suggestion',
         },
         redirects: {
           sheets: redirectTarget,
```
jonathaningram, line 55: To prove that the test works, should we not have the server respond with, say,  a `Sheets2` redirect as well?

jonathaningram (changes requested):

packages/search-ui/src/Input/utils/index.test.ts
jkaho: Right, good point. So we essentially don't support users having different redirects with the same names and different casing because we drop all except the last in the object. Not that I can see anyone doing that, but maybe it's worth mentioning this in the docs?

jonathaningram: Yep or whatever order Object.keys returns its keys.

Worth adding to the docs.

docs/pages/search-ui/input.mdx
```diff
@@ -133,6 +133,12 @@ function Example() {
 }
 ```
 
+### Note
+
+The `Input` component treats redirects as _case-insensitive_, meaning the user-inputted value `Computer` will return the redirect for the query `computer` and vice versa.
+
+The component does not support the handling of multiple redirects for queries of the same word with varied letter casing. For example, if you have created redirects for each of the queries `computer`, `Computer`, and `COMPUTER`, only one of them will be used (dependant upon your browser).
```
jonathaningram, line 140:
```suggestion
The component does not support the handling of multiple redirects for queries of the same word with varied letter casing. For example, if you have created redirects for each of the queries `computer`, `Computer`, and `COMPUTER`, only one of them is used (dependent upon your browser).
```

docs/pages/search-ui/input.mdx
```diff
@@ -133,6 +133,12 @@ function Example() {
 }
 ```
 
+### Note
+
+The `Input` component treats redirects as _case-insensitive_, meaning the user-inputted value `Computer` will return the redirect for the query `computer` and vice versa.
```
jonathaningram, line 138:
```suggestion
The `Input` component treats redirects as _case-insensitive_, meaning the user-inputted value `Computer` returns the redirect for the query `computer` and vice versa.
```

jonathaningram (changes requested):
Small doc nits. Good to go otherwise.

jonathaningram (dismissed):

jkaho:
@jonathaningram sorry needed to rejig the test since it was passing locally but failing in the CI check

jonathaningram (approved):

